PR #20, opened 2020-09-16
Aragon: bump version + add simple example of using Delay app

0xGabi:
The new version fixes the description for the voting aggregator app.


nachomazzara:
@2fd what should we need to do with this or?

Issue #123, opened 2021-09-02
Automated post in forum thread with the results of the votation.


yemel:
Also, it would be great to post a message when the proposal has been marked as Enacted (or other state). The post message should include the comments of the Committee Member.

PR #134, opened 2021-11-02
Automated post in forum thread with the results of the votation

jparnaudo:
Hi! 

This PR addresses https://github.com/decentraland/governance/issues/123


jest-preprocess.js
```diff
@@ -0,0 +1,4 @@
+const babelOptions = {
```
jparnaudo, line 1: added some babel dependencies and configs for using jest with gatsby acording to https://www.gatsbyjs.com/docs/how-to/testing/unit-testing/

src/entities/Proposal/routes.ts
```diff
@@ -1,57 +1,58 @@
...
-import handleAPI, { handleJSON } from 'decentraland-gatsby/dist/entities/Route/handle';
-import validate from 'decentraland-gatsby/dist/entities/Route/validate';
-import schema from 'decentraland-gatsby/dist/entities/Schema'
-import { SNAPSHOT_SPACE, SNAPSHOT_ACCOUNT, SNAPSHOT_ADDRESS, SNAPSHOT_DURATION, signMessage } from '../Snapshot/utils';
-import { Snapshot, SnapshotResult, SnapshotSpace, SnapshotStatus } from '../../api/Snapshot';
-import { Discourse, DiscoursePost } from '../../api/Discourse';
-import { DISCOURSE_AUTH, DISCOURSE_CATEGORY } from '../Discourse/utils';
-import Time from 'decentraland-gatsby/dist/utils/date/Time';
-import ProposalModel from './model';
+import { Request } from "express"
```
jparnaudo, line 1: I'm noticing mi IDE config reorganized importas and changed single quotes for double quotes. I'll change this config to match our codebase

src/entities/Proposal/routes.ts
```diff
@@ -409,11 +410,44 @@ export async function getProposal(req: Request<{ proposal: string }>) {
 }
 
 const updateProposalStatusValidator = schema.compile(updateProposalStatusScheme)
+
+export async function commentProposalUpdateInDiscourse(id: string, caller: string) {
```
jparnaudo, line 414: Some comments about this method:
- this is where the message builder is used, and returns a message depending on the proposal status
- I left some logging so we can test it. For future development I'd like us to consider using a logging library like  https://tslog.js.org/#/ so we can configure the log levels :)
- When an update on discourse fails, for the time being I decided to log it and call it a day. We could implement a retry mechanism

README.md
```diff
@@ -46,12 +50,57 @@ The minimum amount of voting power require to pass a proposal of each type it's
...
 npm run migrate up
 ```
 
+if migrations are not running, or get stuck, please check that you are using node v12
+
+## Run Tests
+
+to run the tests you can do
+```bash
+jest --config jest.config.js --no-watchman --runInBand --colors
```
yemel, line 96: should this replace `npm test` script in `package.json`?

src/entities/Proposal/routes.ts
jparnaudo: - Instead of finding proposals one by one, I'd rather fetch them all at once from the db, with the updated statuses after the job is finished. However, I thought I didn't want to make this PR any longer, so for now it fetches them by one. I hope this is OK.

src/entities/Proposal/discourse-proposal-update-message-builder.ts
```diff
@@ -0,0 +1,58 @@
+import { ProposalAttributes, ProposalStatus } from "./types"
+import { Vote } from "../Votes/types"
+import { calculateResult } from "../Votes/utils"
+
+export class DiscourseProposalUpdateMessageBuilder {
```
yemel, line 5: A shorter file name would be great. Also does this need to be an object?
Maybe a utility function in /entities/Proposal/utils.ts should be enough.

src/entities/Proposal/routes.ts
```diff
@@ -409,11 +410,44 @@ export async function getProposal(req: Request<{ proposal: string }>) {
...
+  const updatedProposal: ProposalAttributes | undefined = await ProposalModel.findOne<ProposalAttributes>({ id })
+
+  if (!!updatedProposal) {
```
yemel, line 417: Consider checking the precondition and return/throw an error right away. Then continue with the logic block.

```
if(!updatedProposal) return console.log(....);
```

src/entities/Proposal/routes.ts
```diff
@@ -409,11 +410,44 @@ export async function getProposal(req: Request<{ proposal: string }>) {
...
+    let message: string
+    try {
```
yemel, line 419: simplify double try catch

test/discourse-proposal-update-message-builder.test.ts
```diff
@@ -0,0 +1,319 @@
+import "jest"
```
yemel, line 1: Tests are located in Entities paths, e.g. `/src/entities/Votes/utils.test.ts`

yemel: Shorter name would be greate

yemel:
Code looks good. Tested on my local environment and works great.

I only found one issue when Enacting/Passing/Rejecting a proposal without entering a text on the modal. It displays "null" in the forum update. Can we check for that case? replacing the null by an empty string will do it.

Screen Shot 2021-11-04 at 13 03 38

yemel:
Looks good to me, @2fd feel free to review and merge/deploy when you have time.

src/entities/Proposal/routes.ts
```diff
@@ -409,6 +410,27 @@ export async function getProposal(req: Request<{ proposal: string }>) {
 }
 
 const updateProposalStatusValidator = schema.compile(updateProposalStatusScheme)
+
+export async function commentProposalUpdateInDiscourse(id: string) {
```
2fd, line 414: use the function `inBackground` to do this

src/entities/Proposal/routes.ts
```diff
@@ -441,6 +463,8 @@ export async function updateProposalStatus(req: WithAuth<Request<{ proposal: str
 
   await ProposalModel.update<ProposalAttributes>(update, { id })
 
+  await commentProposalUpdateInDiscourse(id)
```
2fd, line 466: once you have  this function working with `inBackground` you can remove the `await`

2fd (changes requested):

src/entities/Proposal/routes.ts
jparnaudo: Ty, removed it here and from proposal jobs

jparnaudo:
Tested again for a name ban with the changes requested by @2fd 

image

PR #136, opened 2021-11-09
Add leading options on listings

jparnaudo:
This PR addresses [this issue](https://www.notion.so/Add-Leading-option-on-listings-c8984127244a4ae3b500bb0ac680fc56)

It doesn't show anything if a proposal is active and has no votes
image

When the proposal is active and there is a leading option 
image

When the proposal finishes but the minimum VP isn't met
image

When the proposal finishes with the right amount of VP
image
image

When the proposal finishes with no votes
image

I also tested this with polls and multiple options
image


src/entities/Votes/model.ts
```diff
@@ -27,18 +29,26 @@ export default class VotesModel extends Model<VoteAttributes> {
...
 
   static async createEmpty(proposal_id: string) {
+    const newScore = this.newScoreFor(proposal_id)
+    await this.create(newScore)
+    return newScore
+  }
+
+  static async safelyGetVotesFor(proposal: ProposalAttributes<any>) {
+    let proposalVotes: VoteAttributes | null = await VotesModel.getVotes(proposal.id)
+    return proposalVotes === null || proposalVotes.votes === {} ? await VotesModel.createEmpty(proposal.id) : proposalVotes.votes
```
2fd, line 39: the evaluation `proposalVotes.votes === {}` is always false, this method will create a new vote every time it is call

src/entities/Votes/model.ts
```diff
@@ -27,18 +29,26 @@ export default class VotesModel extends Model<VoteAttributes> {
   }
 
   static async createEmpty(proposal_id: string) {
+    const newScore = this.newScoreFor(proposal_id)
+    await this.create(newScore)
+    return newScore
+  }
+
+  static async safelyGetVotesFor(proposal: ProposalAttributes<any>) {
```
2fd, line 37: some notes:
- we don't use the subfix `for` 
- it is prefer use `proposal_id: string` instead of the whole proposal
- this class is a direct connection with the database, isn't its responsibility know what is safety and what not

src/entities/Votes/model.ts
```diff
@@ -27,18 +29,26 @@ export default class VotesModel extends Model<VoteAttributes> {
...
+    let proposalVotes: VoteAttributes | null = await VotesModel.getVotes(proposal.id)
+    return proposalVotes === null || proposalVotes.votes === {} ? await VotesModel.createEmpty(proposal.id) : proposalVotes.votes
+  }
+
+  static newScoreFor(proposal_id: string) {
```
2fd, line 42: we don't us the subfix `for`

src/entities/Proposal/messages.ts
```diff
@@ -0,0 +1,66 @@
+import { ProposalAttributes, ProposalStatus } from './types'
+import { Vote } from '../Votes/types'
+import { calculateResult } from '../Votes/utils'
+
+export function getUpdateMessage(proposal: ProposalAttributes<any>, votes: Record<string, Vote>) {
```
2fd, line 5: move this to `src/entities/Proposal/template/messages.ts` 🙏

2fd (changes requested):

src/entities/Votes/model.ts
jparnaudo: Moved to Votes/routes

2fd:
Excelente, I'll make de code review soon.

The only things that doesn't feel right is when the leader is positive and the outcome was rejected:



Could confused some users, @yemel WDYT?

yemel:
You are right! Maybe we should just remove the winning text and leave it as:

`Result: VP threshold not met`

2fd (approved):

2fd:
Code seems ok, just the change that yemel suggest

2fd (approved):

src/components/Status/LeadingOption.tsx
```diff
@@ -0,0 +1,30 @@
...
+  function proposalFinished() {
+    return [ProposalStatus.Passed, ProposalStatus.Rejected, ProposalStatus.Finished, ProposalStatus.Enacted].includes(status)
+  }
+
+  return <div className="LeadingOption">
+    {status !== ProposalStatus.Pending &&
+    <Paragraph small secondary>
+      {status === ProposalStatus.Active && (l('page.proposal_detail.leading_option_label'))}
+      {proposalFinished() && (l('page.proposal_detail.finished_result_label'))}{': '}
+      <strong>{leadingOption}</strong>
```
2fd, line 25: Remove `leadingOption` when the output was rejected by vp threshold

2fd (changes requested):

jparnaudo:
Updated labels look like this:

image

image

image

src/components/Status/LeadingOption.tsx
```diff
@@ -0,0 +1,34 @@
...
+export type LeadingOptionProps = React.HTMLAttributes<HTMLDivElement> & {
+  leadingOption: string | null
+  metVP: boolean
+  status: ProposalStatus
+}
+
+export default React.memo(function LeadingOption({ status, leadingOption, metVP }: LeadingOptionProps) {
+  const l = useFormatMessage()
+
+  function proposalFinished() {
```
2fd, line 16: use `useMemo` for this

2fd (changes requested):

2fd (approved):

PR #140, opened 2021-11-25
Discourse comments in proposal page

jparnaudo:
This PR is for [this issue](https://www.notion.so/Add-comments-from-Discourse-to-Proposal-page-9b471e7f3ab744c69473126dc1a50d91)

It introduces a new environment variable, that sets the base url address from which we get the user's forum avatar

`DISCOURSE_BASE_AVATAR_URL=https://sjc6.discourse-cdn.com/standard10`

This is how it looks:

No comments
image

Comments with quotes
image

Comments with images
image


src/entities/Discourse/utils.ts
```diff
@@ -9,4 +10,31 @@ export const DISCOURSE_USER = requiredEnv('DISCOURSE_USER')
 export const DISCOURSE_AUTH: DiscourseAuth = {
   apiKey: DISCOURSE_API_KEY,
   apiUsername: DISCOURSE_USER
-}
\ No newline at end of file
+}
+export const BASE_AVATAR_URL = requiredEnv('DISCOURSE_BASE_AVATAR_URL')
+const DEFAULT_AVATAR_SIZE = '45'
+
+function setAvatarSize(avatar_url: string) {
```
2fd, line 17: `set` is not a proper prefix for a function that doesn't modify anything, `get` is preferred

src/entities/Proposal/routes.ts
```diff
@@ -495,6 +496,17 @@ export async function removeProposal(req: WithAuth<Request<{ proposal: string }>
   return true
 }
 
+export async function proposalComments(req: Request<{ proposal: string }>){
+  const proposal = await getProposal(req)
+  try{
+    const comments = await Discourse.get().getTopic(proposal.discourse_topic_id)
+    return filterComments(comments);
+  } catch (e) {
+    console.log(`Could not get proposal comments, `, e)
```
2fd, line 505: the use of `Logger` is preferred over `console`

```suggestion
    logger.error(`Could not get proposal comments`, e)
```

2fd (changes requested):

src/entities/Proposal/routes.ts
jparnaudo: Yeah, I'd rather use a logger but I didn't see it anywhere in the code, so I went with `console` for being consistent with what's already in the router. 

Maybe we should consider creating a separate PR for changing all the `console.log` uses to `logger.error` / `logger.info`?

jparnaudo: I created a new [ticket in notion](https://www.notion.so/Replace-console-logs-for-logger-8dee70749e23468391180023b980ef83) for replacing the console.log uses with Logger

2fd (approved):

PR #143, opened 2021-11-29
Suscribe to dao updates via email

jparnaudo:
This PR addresses [this issue](https://www.notion.so/Suscribe-to-DAO-updates-via-email-4b18e748a0a043cf82dd701b7ac55815)

It introduces new env varible, to subscribe to decentraland's mailchimp.
`GATSBY_DECENTRALAND_API: https://decentraland.org`

Banner only shows for logged users
image

On every proposal category
image

Looks like this
image

Has local email format validation
image

Displays mailchimp's format/data validation errors
image

Shows success modal after successful subscription
image

Saves subscribed user address in local storage. 

If we log out, banner does not appear
image

If we log with another user, it shows again:
image


2fd:
Isn't that too intrusive for users. Cloud be at least add a close option and use `localStorage` to define if a user has subscribed or closed the banner and hide it?

yemel:
I have some comments on this:

* The newsletter should be visible always, not only for logged users.
* The header should be only visible on "All Proposals". If you filter by any category (POI, POLL, GRANT, etc) the newsletter header should be hidden.

Since it will be only visible at "All Proposals" it won't be too intrusive. I don't think we need a way to dismiss it.

jparnaudo:
Made the requested changes. 

Now the banner also shows for anonymous users (until they subscribe)

til

If a non subscripted user signs in, despite an anonymous user subscripted, banner for subscription shows:

user1

If a different unsubscripted user logs in, banner shows again.  When subscribed, it disappears.

user2

If there's no anonymous user subscribed, and subscribed user logs in, banner disappears. When user logs out, banner shows again.

test3

src/api/Governance.ts
```diff
@@ -206,6 +207,16 @@ export class Governance extends API {
     return result.data
   }
 
+  async subscribeToNewsletter(email: string) {
```
2fd, line 210: This featue shouldn't need and extra endpoint, can use the `/subscribe` endpoint directly

src/components/Modal/NewsletterSubscriptionModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export function NewsletterSubscriptionModal({ onSubscriptionSuccess, subscribed, ...props }: NewsletterSubscriptionModalProps) {
+  const [account] = useAuthContext()
+  const l = useFormatMessage()
+  const [isValid, setIsValid] = useState(true);
+  const [message, setMessage] = useState('');
+  const [email, setEmail] = useState('')
+
+  const emailRegex = /\S+@\S+\.\S+/;
```
2fd, line 35: If you have a variable that never changes should be defined outside the component.

This case in particular it is preferred to use:
```ts
import isEmail from 'validator/lib/isEmail'
```

src/components/Modal/NewsletterSubscriptionModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+      setMessage('');
+      setEmail(email)
+    } else {
+      setIsValid(false);
+      setMessage(l('modal.newsletter_subscription.email_error_message') || '');
+      setEmail(email)
+    }
+  };
+
+  function saveSubscription() {
```
2fd, line 50: use `useCallback`

src/components/Modal/NewsletterSubscriptionModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const subscriptions: string[] = JSON.parse(localStorage.getItem(NEWSLETTER_SUBSCRIPTION_KEY) || '[]');
+    subscriptions.push(account || ANONYMOUS_USR_SUBSCRIPTION)
+    localStorage.setItem(NEWSLETTER_SUBSCRIPTION_KEY, JSON.stringify(subscriptions))
+  }
+
+  function resetModal() {
```
2fd, line 56: use  `useCallback`

src/components/Modal/NewsletterSubscriptionModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const validateEmail = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const email = event.target.value;
+    if (emailRegex.test(email)) {
+      setIsValid(true);
+      setMessage('');
+      setEmail(email)
+    } else {
+      setIsValid(false);
+      setMessage(l('modal.newsletter_subscription.email_error_message') || '');
+      setEmail(email)
```
2fd, line 46: if you are setting more than one variable at same time a composite state is preferred:

```ts
const [ state, setState ] = useState<{ isValid, message, email }>({ isValid: true, message: '', email: ''})

// ...

setState({
  isValid: true,
  message: '',
  email: email
})
```

src/entities/NewsletterSubscription/routes.ts
```diff
@@ -0,0 +1,26 @@
+import { Request } from 'express'
+import routes from 'decentraland-gatsby/dist/entities/Route/routes';
+import handleAPI from 'decentraland-gatsby/dist/entities/Route/handle';
+import { Decentraland } from '../../api/Decentraland'
+import { NewsletterSubscriptionResult } from './types'
+
+export default routes((route) => {
+  route.post('/newsletter/:email', handleAPI(subscribeToNewsletter))
```
2fd, line 8: plesae use the endpoint directly from the front end

2fd (changes requested):

src/api/Decentraland.ts
```diff
@@ -0,0 +1,32 @@
...
+  static get() {
+    return this.from(this.Url)
+  }
+
+  async subscribe(email: string){
+    return this.fetch(
+      '/subscribe',
+      this.options()
+        .method('POST')
+        .json({ email, interest: "dao"})
```
2fd, line 29: sorry guy, we need yo change this to: 

```suggestion
        .json({ email, list: "governance"})
```

2fd (changes requested):

src/api/Decentraland.ts
jparnaudo: done!

2fd (approved):

Issue #142, opened 2021-11-28
Fix: package-lock.json wrong dependency path 


oliviel:
The package-lock.json is facing a problem finding a dependency after the merge of [Automated post in the forum thread with the results of the votation](https://github.com/decentraland/governance/commit/9dca8654765086f4530692f4d2ec315fc9c7b6f5) because of that you cant install all the dependencies of the package. The problem is with the dependency of  [ethereumjs-abi@git+https://github.com/ethereumjs/ethereumjs-abi.git](https://github.com/decentraland/governance/commit/9dca8654765086f4530692f4d2ec315fc9c7b6f5#r61003004) inside the package-lock.json without this fixed anyone can't install the dependencies of the project and start contributing.

I tested this on a mac and ubuntu and is displaying the same error as the image below

<img width="1115" alt="Screen Shot 2021-11-28 at 2 30 05 PM" src="https://user-images.githubusercontent.com/16074043/143781183-5edef342-73f9-4e6e-ae06-d97962a2d3a7.png">

yemel:
@jparnaudo could you check this out?

PR #144, opened 2021-12-01
Fix dependency path of package-lock.json

oliviel:
This PR implements:

- Fix the dependency path of ethereumjs-abi.
- Enable the project to install all the dependencies without issues.

This PR resolves issue #142


2fd:
That is only a problem when you use `npm > 6`. Please use the correct version. (An update is comming)
If the problem persist use the following command `git config --global url."https://github.com/".insteadOf "ssh://git@github.com/"`

Issue #116, opened 2021-08-20
Add sorting to proposals


schallock:

yemel:
<img width="810" alt="Screen Shot 2021-11-23 at 15 44 04" src="https://user-images.githubusercontent.com/279449/143084965-e5477b7e-b3c5-4b1a-a2aa-fe86950ae15b.png">

Allow the sorting of proposals by LATEST and PARTICIPATION. Latest sorts by date and Participation by total VP that voted.

ginocingolani:
Solved by https://github.com/decentraland/governance/pull/147

PR #147, opened 2021-12-06
Add sorting to proposals

oliviel:
This PR implements:

- add total_vp field to the proposals table and migration.
- create a subscription method to update the total_vp of votes each minute.
- Create a SortingMenu component to filter in the front end to sort by latest (created_at) and participation (total_vp).

This PR resolves issue #116


2fd:
Two thing:
- `package-lock.json` should be modified that way, check your previous PR for more details
- you are sorting list on front (and on each render) which is not a proper approach, let discuss this with more details

PR #156, opened 2021-12-29
Transparency page mission

jparnaudo:
This PR addresses the first part of the [transparency page](https://www.notion.so/Add-Transparency-Page-f2e361c5192044a59948609f017cfb4a)

- This is how it looks
image

- Transparency page shows even if user is not logged.
- New balances are created only if the value differs from the last stored value
- It introduces two new env vars for configuring block explorers base urls
- I created methods in the *Model classes that can be mocked with jest, to emulate the db

Caveats:
- Frontend is not responsive yet. Browser version looks good. Phone version doesn't.
- There's a repeated function `formattedTokenBalance`, that I could not import from either component that needs it, so I ended up redefining it in each component. Any advice on how to avoid this would be great.

Cannot assign reviewers yet, so I'll tag them: @yemel @2fd


src/components/Token/TokenBalanceCard.tsx
```diff
@@ -0,0 +1,32 @@
...
+import './TokenBalanceCard.css'
+import Icon from 'react-crypto-icons'
+import TokensPerWalletPopup from './TokensPerWalletPopup'
+import { AggregatedTokenBalance, TokenBalance } from '../../entities/Balance/types'
+
+export type TokenBalanceCardProps = React.HTMLAttributes<HTMLDivElement> & {
+  aggregatedTokenBalance: AggregatedTokenBalance
+}
+
+function formattedTokenBalance(tokenBalance: TokenBalance) {
```
jparnaudo, line 13: I could not import this function from '../../entities/Balance/utils', not sure why.

yemel: It's most probably a circular dependency error. Try putting it in Wallet/utils.ts

.env.example
```diff
@@ -52,3 +52,7 @@ ALCHEMY_API_KEY=
 
 # Discord
 GATSBY_JOIN_DISCORD_URL=https://dcl.gg/discord
+
+# Block Explorers
+ETHERSCAN_BASE_URL=https://etherscan.io/
```
yemel, line 57: I believe it's not necessary to configure this at this level. If other project forks this repo they will probably still use the same block explorer for ETH and MATIC.

jparnaudo:
Update:

- Responsive layout
- Token balances are all clickable and display balances by wallet
- The whole wallet balance row is now clickable
- Audit on block explorer link always displays on mobile
- Wallet balances with 0 amount do not show
- Total token balances are sorted by amount
- Check our docs button
- Removed block explorers env vars

package.json
```diff
@@ -13,6 +13,7 @@
     "clean": "gatsby clean",
     "version": "tsc -p .",
     "migrate": "DOTENV_CONFIG_PATH=.env.development ts-node -r dotenv/config.js ./node_modules/node-pg-migrate/bin/node-pg-migrate -j ts -m src/migrations -d CONNECTION_STRING",
+    "create": "DOTENV_CONFIG_PATH=.env.development ts-node -r dotenv/config.js ./node_modules/node-pg-migrate/bin/node-pg-migrate -j ts -m src/migrations create",
```
2fd, line 16: Please remove this, should use `npm run migrate create` instead

src/api/Alchemy.ts
```diff
@@ -0,0 +1,94 @@
...
+        return this.from('https://eth-mainnet.alchemyapi.io')
+      case ChainId.MATIC_MAINNET:
+        return this.from('https://polygon-mainnet.g.alchemy.com')
+    }
+    throw new Error('Unsuported chain id')
+  }
+
+  async getTokenBalances<P extends keyof TokenBalancesResponse>(address: string, tokens: string[]) {
+    return await this.fetch<TokenBalancesResponse>(
+      '/v2/HcW-dJIcedLgGAB8htL7M1L8ir2Sihfe',
```
2fd, line 55: isn't `HcW-dJIcedLgGAB8htL7M1L8ir2Sihfe` and apikey? if it is anyone cloud copy it and outlimit our requests

wouldn't be better use a server request for that?

(I'm asking cause I don't know how alchemy works)

src/entities/Balance/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+  const blockNumber = await Alchemy.get(chainId).getBlockNumber()
+  return blockNumber.result
+}
+
+export async function latestConsistentBlockNumber(chainId: ChainId) {
+  let currentBlockNumber = await getBlockNumber(chainId)
+  return (asNumber(currentBlockNumber) - asNumber(0x100)).toString(16)
+}
+
+export async function getBalances() {
```
2fd, line 61: As this function can update the database or even return nothing, I'd prefer a name like `getUpdateBalances`

src/entities/Balance/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+  const wallets: WalletAttributes[] = await WalletModel.find()
+
+  for (const wallet of wallets) {
+    const tokenBalances: TokenBalanceResponse[] = await getTokenBalancesFor(wallet)
+    for (const tokenBalance of tokenBalances) {
+      try {
+        const token = await TokenModel.findMatchingToken(wallet.network, tokenBalance.contractAddress)
+        const previousBalance = await BalanceModel.findLatest(wallet.id, token.id)
+        if (previousBalance && previousBalance.amount === tokenBalance.tokenBalance) {
+          return
```
2fd, line 74: would't this ends the execution of the function and return `undefined` if found only one balance that match the previous one? seems like an unexpected side-effecs

src/entities/Balance/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+  const createdBalances = []
+  const errors = []
+
+  const wallets: WalletAttributes[] = await WalletModel.find()
+
+  for (const wallet of wallets) {
+    const tokenBalances: TokenBalanceResponse[] = await getTokenBalancesFor(wallet)
+    for (const tokenBalance of tokenBalances) {
+      try {
+        const token = await TokenModel.findMatchingToken(wallet.network, tokenBalance.contractAddress)
```
2fd, line 71: I'm worry about having a query inside a `for`. Wouldn't be easier get all Tokens at the beginning and make the cartesian in memory (or use a `Map`) you will use only one query and get the same amount of iterations.

src/entities/Balance/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+        const previousBalance = await BalanceModel.findLatest(wallet.id, token.id)
```
2fd, line 72: Same here, Isn't there a way to get this result for a list of token/wallet?

src/entities/Balance/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+        symbol: token.symbol!,
+        amount: toPaddedHexString(0),
+        decimals: token.decimals!
+      },
+      tokenInWallets: []
+    })
+  }
+
+  for (const balance of latestBalances) {
+    const token: TokenAttributes = await TokenModel.findToken(balance.token_id)
```
2fd, line 132: same here, it would be better to query the database just once with the list of tokens

src/entities/Balance/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+    const wallet = await WalletModel.findWallet(balance.wallet_id)
```
2fd, line 133: same here,  it would be better to query the database just once with the list of wallet

src/migrations/1639686932772_create-balances-table.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    amount: {
+      type: 'TEXT',
+      notNull: true
+    },
+    created_at: {
+      type: 'TIMESTAMPTZ',
+      notNull: true
+    }
+  })
```
2fd, line 30: This table needs indexes, `tokens` and `wallets` are small and fixed size, but this one grows indefinitely, It will be a problem in the future

2fd (changes requested):

package.json
jparnaudo: ty, that's what I needed

jparnaudo:
@2fd thank's for your thorough review, always much appreciated!

I closed this PR because @yemel suggested taking a different direction for the transparency page, regarding on how we get the info. I'm going to reuse a part of it in a future PR.

Issue #89, opened 2021-05-28
[FORM] Remove a POI


2fd:
Add a way to submit a proposal for removing a Point of Interest.

* A modal window opens on the submit new proposal page.
* The user completes a form to indicate the X, Y coordinates and reasons to remove it:
  * We need to validate if the POI coordinates are valid and if the POI is listed in the smart contract.
  * The reason is mandatory, markdown format should be supported (check grant's form)
* The proposal should be created under the "POI" category using the corresponding VP threshold and voting period.

Check out the [figma file](https://www.figma.com/file/e0IFJjplUdYdQJ2NtLwvMf/DAO-v2-(Guane)?node-id=1796%3A497) for UI references.
  
![new-proposal - poi](https://user-images.githubusercontent.com/279449/148405396-fca66b08-6fa5-4104-b9ac-994310ef6df2.jpg)

![proposal form_ revoke poi](https://user-images.githubusercontent.com/279449/148405416-282eeefc-b8bf-47a1-a214-dec2373c8f62.jpg)

![proposal detail_ revoke poi](https://user-images.githubusercontent.com/279449/148405436-0973e515-2c91-4e7e-8944-0b68ccf2ad00.jpg)

PR #157, opened 2022-01-10
feat: Remove a POI (#89)

ncomerci:
* Refactor POI page
* Added modal to choose 'Add' or 'Remove'
* Refactor '/proposals/poi' API endpoint


src/components/Category/CategoryBanner.tsx
```diff
@@ -8,26 +8,35 @@ import TokenList from 'decentraland-gatsby/dist/utils/dom/TokenList'
...
+  [ProposalType.POI]: require("../../images/icons/poi.svg"),
+  [ProposalType.AddPOI]: require("../../images/icons/add-poi.svg"),
+  [ProposalType.RemovePOI]: require("../../images/icons/remove-poi.svg"),
+  [ProposalType.BanName]: require("../../images/icons/ban-name.svg"),
+  [ProposalType.Grant]: require("../../images/icons/grant.svg"),
+  [ProposalType.Poll]: require("../../images/icons/poll.svg"),
 }
 
-export type CategoryBannerProps = Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'children'> & {
+export type CategoryBannerProps = Omit<
```
jparnaudo, line 20: Please disable the auto-formatting so it's easier for reviewers to tell which lines are edited 😄

src/components/Category/CategoryBanner.tsx
```diff
@@ -8,26 +8,35 @@ import TokenList from 'decentraland-gatsby/dist/utils/dom/TokenList'
...
 
 const icons = {
-  [ProposalType.Catalyst]: require('../../images/icons/catalyst.svg'),
-  [ProposalType.POI]: require('../../images/icons/poi.svg'),
-  [ProposalType.BanName]: require('../../images/icons/ban-name.svg'),
-  [ProposalType.Grant]: require('../../images/icons/grant.svg'),
-  [ProposalType.Poll]: require('../../images/icons/poll.svg'),
+  [ProposalType.Catalyst]: require("../../images/icons/catalyst.svg"),
+  [ProposalType.POI]: require("../../images/icons/poi.svg"),
+  [ProposalType.AddPOI]: require("../../images/icons/add-poi.svg"),
```
jparnaudo, line 13: Since the proposals add/remove poi should still be  a `ProposalType.POI` , I don't think adding new proposal type is necessary. 

Maybe what we are missing here is a new `PoiType` type, that could be the key for this Icons?

src/entities/Proposal/routes.ts
```diff
@@ -188,22 +189,15 @@ export async function createProposalBanName(req: WithAuth) {
 }
 
 const newProposalPOIValidator = schema.compile(newProposalPOIScheme)
-export async function createProposalPOI(req: WithAuth) {
+type VerifyFunction = (config: NewProposalPOI) => Promise<void>;
```
jparnaudo, line 192: If we change `NewProposalPoI` to have a `poiType: PoiType`, then we could use just one endpoint and validate accordingly. (and also determine how to create the proposal template)

src/entities/Proposal/routes.ts
```diff
@@ -367,6 +390,13 @@ export async function createProposal(data: Pick<ProposalAttributes, 'type' | 'us
   //
   // Create proposal in DB
   //
+  switch (data.type) {
```
jparnaudo, line 393: This switch for me is the hint that we really don't need new ProposalTypes 😄

src/entities/Proposal/templates/addPoi.ts
```diff
@@ -1,7 +1,10 @@
+import useFormatMessage from "decentraland-gatsby/dist/hooks/useFormatMessage";
 import Land from "decentraland-gatsby/dist/utils/api/Land";
 import { NewProposalPOI } from "../types";
 import { formatMarkdown, template } from "./utils";
 
+// const l = useFormatMessage();
+
 export const title = (proposal: NewProposalPOI) =>
```
jparnaudo, line 8: This is where a the `poiType` field in the `NewProposalPOI` would also be useful. the `addPoi.ts` file could contain all the logic needed for generating either template

src/entities/Proposal/templates/index.ts
```diff
@@ -18,7 +19,11 @@ type NewConfiguration =
 export const title = async ({ type, configuration }: { type: ProposalType, configuration: NewConfiguration }) => {
   switch (type) {
     case ProposalType.POI:
-      return poi.title(configuration as any)
+      return "" // no title
```
jparnaudo, line 22: all this new cases wouldn't be needed because the `title`/`description`/`etc`/ fn inside addPoi can determine what to do based on the `NewProposalPoi` object

src/pages/submit/remove-poi.tsx
```diff
@@ -0,0 +1,257 @@
...
+} from "../../components/Layout/ContentLayout"
+import { Governance } from "../../api/Governance"
+import locations from "../../modules/locations"
+import loader from "../../modules/loader"
+import useAuthContext from "decentraland-gatsby/dist/context/Auth/useAuthContext"
+import Head from "decentraland-gatsby/dist/components/Head/Head"
+import MarkdownNotice from "../../components/Form/MarkdownNotice"
+import "./submit.css"
+
+type POIState = {
```
jparnaudo, line 34: this is already defined in add-poi

src/pages/submit/remove-poi.tsx
```diff
@@ -0,0 +1,257 @@
...
+}
+
+const edit = (state: POIState, props: Partial<POIState>) => {
+  return {
+    ...state,
+    ...props,
+  }
+}
+
+function assertInMap(value: string | number) {
```
jparnaudo, line 54: also defined in `add-poi`.

src/pages/submit/remove-poi.tsx
```diff
@@ -0,0 +1,257 @@
...
+      Promise.resolve()
+        .then(async () => {
+          const x = asNumber(state.value.x)
+          const y = asNumber(state.value.y)
+
+          if (!(await isAlreadyPointOfInterest(x, y))) {
+            throw new Error(`error.poi.coordinates_are_not_a_poi`)
+          }
+
+          return Governance.get().createRemoveProposalPOI({
```
jparnaudo, line 118: All this validations and fields in common make me think if we should have only one component for adding/removing, and depending on a url param, change the displayed text and behavior 

For example, depending on the type we could change the validations (as in the back) and do

```js
return Governance.get().createNewProposalPoi({
  x,
  y,
  description: state.value.description,
  poiType: poiType
 })
```
What do you think @yemel ?

src/entities/Proposal/routes.ts
```diff
@@ -213,6 +207,35 @@ export async function createProposalPOI(req: WithAuth) {
   })
 }
 
+export async function createAddProposalPOI(req: WithAuth) {
+
+  const verify = async (config: NewProposalPOI) => {
```
jparnaudo, line 212: I like the idea of a verifying fn depending on the poi type.

Maybe if we decide to use a PoiType, it could decide how to validate internally.

jparnaudo (changes requested):
I like it in general, and left some ideas that can be helpful! 😸

src/pages/submit/remove-poi.tsx
ncomerci: Thank you very much for all the feedback! Yes, before submitting the PR, I considered refactoring the POI page to a component given the amount of shared functionality. But I was a bit intimidated by the amount of code in the file and I was afraid of breaking something just when I was finishing.

Taking your suggestions, I'll make the changes and resubmit the PR.

Thanks again!

src/entities/Proposal/templates/addPoi.ts
ncomerci: The comment here is because I tried to replace the hardcoded strings in the template (to be consistent with i18n), but when I added that line I got a css error that didn't make any kind of sense and I didn't know how to fix it. I left the comment to ask you later.

yemel:
LGTM

* Tested locally and works as expected.
* Code reviewed and no issues found.

PR #158, opened 2022-01-10
Three stage proposal flow

jparnaudo:
WIP


src/api/Governance.ts
```diff
@@ -210,4 +223,9 @@ export class Governance extends API {
     const result = await this.fetch<ApiResponse<ProposalComment[]>>(`/proposals/${proposal_id}/comments`)
     return result.data
   }
+
+  async getPassedProposals(proposal_type: ProposalType):Promise<DropdownItemProps[]> {
```
2fd, line 227: This api shouldn't know anything about the front end.

If you need change the shape of an api response please do it i front end logic, don't create endpoints to get each formatted data the front-end needs

src/api/Snapshot.ts
```diff
@@ -263,4 +263,15 @@ export class Snapshot extends API {
...
     return this.getScores(info.id, info.strategies, info.network, addresses)
   }
-}
\ No newline at end of file
+
+  async getVotingPower(address?: string | null, space?: string | null){
+    if (!address || !space) {
+      return 0
+    }
+    const info = await Snapshot.get().getSpace(space)
```
2fd, line 271: you are inside the instance, here you can use `this.getSpace(space)`

src/entities/Proposal/model.ts
```diff
@@ -162,4 +162,27 @@ export default class ProposalModel extends Model<ProposalAttributes> {
 
     return proposals.map(this.parse)
   }
-}
\ No newline at end of file
+
+  static async getPassedProposals(proposal_type: ProposalType) {
```
2fd, line 166: please don't add custom queries if you don't need it for performance reasons.

is preferible to use the old endpoint so we can apply cache strategies if it is required

src/entities/Proposal/routes.ts
```diff
@@ -520,6 +574,37 @@ export async function proposalComments(req: Request<{ proposal: string }>){
...
+    type: expectedProposalType,
+    deleted: false
+  })
+  if (!linkedProposal) {
+    throw new RequestError(`Could not find a matching ${expectedProposalType} proposal for "${linkedProposalId}"`, RequestError.NotFound)
+  }
+}
+
+async function validateSubmissionThreshold(user: string, submissionThreshold?: string) {
+  const requiredVp = Number(submissionThreshold || requiredEnv('SUBMISSION_THRESHOLD_POLL'))
```
2fd, line 601: the function `requiredEnv` should be use on top level, its goal is to know at bootstrap time if there is a missing ENV

2fd (changes requested):

src/api/Governance.ts
```diff
@@ -2,6 +2,7 @@ import API from 'decentraland-gatsby/dist/utils/api/API';
 import { ApiResponse } from 'decentraland-gatsby/dist/utils/api/types';
 import Time from 'decentraland-gatsby/dist/utils/date/Time';
 import env from 'decentraland-gatsby/dist/utils/env';
+import { DropdownItemProps } from 'semantic-ui-react'
```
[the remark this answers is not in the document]
jparnaudo: Ty, replaced this with `import type`

src/api/Governance.ts
jparnaudo: Ok, I get what you are saying.

I created this endpoint because we don't need the whole proposal data in the front end, just a list of title and an id. 
Sending all that info to the frontend seemed a little bit over the top for populating a dropdown.

And since Governance and Draft proposal submission pages both have this kind of dropdown, it seemed worthy to have an endpoint for it.

Maybe we can find another solution that does not return the data formatted as a dropdown item (so it's agnostic to the frontend needs) but still returns a proposal list with just the title and id.

What are your thoughts?

jparnaudo: maybe something like 

```js
async getProposalsList(proposal_type: ProposalType, proposal_status: ProposalStatus): Promise<Pick<ProposalAttributes, 'id' | 'title'>>{
...
}
```

??

2fd: What about add and extra parameter to the current api `properties`, so you can use an `omit` function over the result

```text
GET /api/proposal?status=passed&type=poi&properties=id,title

{
    "success": true,
    "data": [
        {
            "id": "...",
            "title": "..."
        }
    ]
}

```

you just need to add and extra prop to `getProposals` on the api

```typescript
export type GetProposalsFilter = {
  // ...
  properties: (keyof ProposalAttributes)[] 
}

Governance.get().getProposals({ type: 'poi', status: 'passed', properties: ['id', title]  })
```

src/api/Governance.ts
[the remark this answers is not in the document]
2fd: Sorry, I expressed myself wrong, my concern was about importing from a front-end library, this api lib shouldn't know anything about the front-end

yemel:
Please create a new PR with V2.

PR #160, opened 2022-01-21
Three stage proposal flow v1

jparnaudo:
This PR addresses [this issue](https://www.notion.so/dcl-dao/Three-Stage-Proposal-Flow-187d2063e5854fe2bca58409c42a8e19)

The only difference is that proposals submissions are limited to promotions, and proposal selector dropdown is disabled.


.env.example
```diff
@@ -47,6 +50,16 @@ GATSBY_DURATION_GRANT_TIER4 = 1200
 GATSBY_DURATION_GRANT_TIER5 = 1200
 GATSBY_DURATION_GRANT_TIER6 = 1200
 
+# Duration for governance process proposals
+GATSBY_DURATION_PRE_PROPOSAL_POLL=1200
```
yemel, line 54: Make it shorter and consistent:
GATSBY_DURATION_POLL
GATSBY_DURATION_DRAFT
GATSBY_DURATION_GOVERNANCE

This way it matches with the naming of the threshold. GATSBY_SUBMISSION_THRESHOLD_POLL/DRAFT/GOVERNANCE

src/entities/Proposal/model.ts
```diff
@@ -162,4 +162,15 @@ export default class ProposalModel extends Model<ProposalAttributes> {
 
     return proposals.map(this.parse)
   }
-}
\ No newline at end of file
+
+  static async getTitle(proposal_id: string) {
```
yemel, line 166: Is this necessary? I think you can do the query like this:

`const proposal = await ProposalModel.findOne<ProposalAttributes>({ id, deleted: false })`

PR #171, opened 2022-01-27
Align labels

jparnaudo:
This is how it looked
image

With the fix
image


yemel:
The See Votes modal has an error. It only displays Yes/No options, it should honor the real names of choices!

jparnaudo:
Added the real name of the choices, and made more room for option names:

Several options:
image

This is how a long option looked
image

How it looks now
image

Mobile
image

image

image

image

yemel (approved):
LGTM, tested locally 👍

2fd (approved):

PR #174, opened 2022-02-03
Feat: navigation proposals

ncomerci:
This PR addresses [this issue](https://www.notion.so/dcl-dao/Navigate-through-proposals-ea8cb71c47c44b94a60ba5a3d90b4381)

**IMPORTANT**
- Now if a proposal id is passed to the hook "useProposals", it will return that proposal, the previous one and the next one (if they exist) taking into account the filters

## Desktop view

image
image
image

## Mobile vew

image

### Proposal titles with more than 100 characters
image


ncomerci:
Given the way the back button functionality is designed, if you navigate back and forth, when you use the back button, it respects the navigation you have done.

Maybe it should go directly back to the main page but I didn't modify it because I wasn't sure what the expected behavior was.

image

yemel:
@ncomerci you are right about the back behavior. Please change it to take you to the listing page using the corresponding filters.

yemel:
Tested locally and works as expected. I'm interested in @2fd review about the sql queries.
Pending revision of the back button in the proposal page.

yemel (changes requested):
Please review the back button behaviour.

src/components/Proposal/ProposalNavigation.tsx
```diff
@@ -0,0 +1,77 @@
...
+import ProposalNavigationItem, {NavigationType} from './ProposalNavigationItem';
+import { UrlParamsContext } from '../Context/UrlParamsContext';
+import useProposals from '../../hooks/useProposals';
+import { useLocation } from '@reach/router'
+import './ProposalNavigation.css';
+
+function ProposalNavigation() {
+
+  const location = useLocation()
+  const p = useMemo(() => new URLSearchParams(location.search), [ location.search ])
```
jparnaudo, line 12: Super minor comment: In general I'd rather have more descriptive names, even if it's an index or something that isn't of our domain

jparnaudo (approved):

ncomerci:
> Please review the back button behavior.

@yemel back button behavior modified

src/components/Proposal/ProposalNavigation.tsx
ncomerci: fixed!

ginocingolani:
We closed this one since we weren't ready to move with this feature forward and the PR has been open for years.

Issue #173, opened 2022-01-28
Include Vesting Contract on Enacted Page


jparnaudo:
**Context**
When a grant is enacted by a DAO Committee member, he usually has to put the vesting contract address in the comment.

We need to add a separate field for the vesting contract in the modal:

![image](https://user-images.githubusercontent.com/2858950/151581960-b5f1148e-8fa6-4f73-a117-2523be4f78f5.png)

**Definition**

- Add database migration to include new field on proposals: vesting_address
- Modify confirmation screen when enacting a proposal (by DAO Committee member)
- Vesting contract is not mandatory. It should validate it is an address (0x....).
- The update should post in the forum thread the vesting contract and the comments.
- Create a “seed migration” that populates the fields for past vesting contracts (check data here: [https://docs.google.com/spreadsheets/d/1vXIf2IzJt83KL5VuGhVySRGY9064kUX-FAEB9TDP-kw/edit?usp=sharing](https://docs.google.com/spreadsheets/d/1vXIf2IzJt83KL5VuGhVySRGY9064kUX-FAEB9TDP-kw/edit?usp=sharing)) - Ask @Yemel Jardi if you have questions.

edit: this is only for Grant proposals!

PR #179, opened 2022-02-08
173 - Include Vesting Contract on Enacted Page

matiibai:
### Include Vesting Contract on Enacted Page
[Original Issue](https://github.com/decentraland/governance/issues/173)

Changes:
- chore: database migration to include vesting_address field in proposals
- feat(Front): Proposal enacting has an optional vesting contract field that is sent to the server
- feat(Server): Update proposal status endpoint saves vesting address and post it to forum
- chore: database migration using previous contracts to update them to use vesting_address

The final result is the following:

image

This new field is optional, and in case it is completed, it must be valid in order to succeed.

image

The updates regarding the spreadsheet were generated using a Python script, so this way we just have the queries we need and there is no need to read from the file system any file or add any dependency to parse csv files or documents alike. In case you prefer that way, please do tell me. I decided to go with what I considered it to be the lightweight solution.

This was the script used:
```python
import pandas as pd
import json

FILENAME = "grant_payouts_all_vestings"

df = pd.read_csv(FILENAME + ".csv", delimiter=",")
df1 = df.where(pd.notnull(df), None)
res = df1.to_dict('records')

sql = ''

for rec in res:
    id = rec['ID']
    vesting = rec['Vesting Address']
    q = ''
    if vesting is None: # null
        q += f"UPDATE proposals SET vesting_address = NULL WHERE id = '{id}';"
    elif vesting == 'rejected': # rejected
        q += f"UPDATE proposals SET vesting_address = NULL, status = 'rejected' WHERE id = '{id}';"
    else: # valid address
        q += f"UPDATE proposals SET vesting_address = '{vesting}' WHERE id = '{id}';"
    sql += q + '\n'

print(sql)
```

In case something is missing or catches your attention, please tell me so I can fix it for you.

### DIFF
--- a/src/api/Governance.ts
+++ b/src/api/Governance.ts
@@ -144,13 +144,13 @@ export class Governance extends API {
     return result.data
   }
 
-  async updateProposalStatus(proposal_id: string, status: ProposalStatus, description: string | null = null) {
+  async updateProposalStatus(proposal_id: string, status: ProposalStatus, vesting_address: string | null, description: string | null = null) {
     const result = await this.fetch<ApiResponse<ProposalAttributes>>(
       `/proposals/${proposal_id}`,
       this.options()
         .method('PATCH')
         .authorization()
-        .json({ status, description })
+        .json({ status, vesting_address, description })
     )
 
     return result.data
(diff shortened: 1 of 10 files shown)


yemel:
LGTM, tested locally:

* DB migration and seed data are correct.
* Update proposal status modal works as expected.
* Messages on forum includes the vesting address and comments if present.

yemel (approved):

2fd (approved):

PR #180, opened 2022-02-09
Filter emojis from image styling

jparnaudo:
Before fix
image

image

After fix

image


yemel (approved):
LGTM, tested locally.

PR #186, opened 2022-02-16
Proposals Search

1emu:
This PR addresses [this issue](https://www.notion.so/dcl-dao/Search-Experience-on-Governance-dApp-46f3cec09059472e867ac57fd740f427)

Done
- [X] search api 
- [X] filter by timeFrame
- [X] order by creation time
- [X] animated collapsible filters & search input
- [X] burger menu search for mobile
- [x] Issue when loading the mobile version with a string query, where the proposals section translates more than it should, because it's assuming that filters are not collapsed.

### Search view on Desktop 

Proposals page
image

Active search
image

No matches
image

Search input is displayed for all pages
image

### Search view on Mobile

Closed - can order by creation date
image

image

Open Menu - no search
image

Active search - filters closed
image

Active Search - filters opened
image

Active Search with filters used and displayed
image

Active Search with filters used and collapsed
image

Active Search with filters used and closed
image


src/entities/Proposal/model.ts
```diff
@@ -146,20 +191,57 @@ export default class ProposalModel extends Model<ProposalAttributes> {
       return []
     }
 
+    let timeFrame = this.parseTimeframe(filter.timeFrame)
+
+    let orderBy = 'p.created_at'
+    if (filter.search) {
+      orderBy = '"rank"'
+    }
+    let orderDirection = 'DESC'
```
andyesp, line 200: You can write this `const orderDirection = filter.order || 'DESC'`

src/entities/Proposal/model.ts
```diff
@@ -146,20 +191,57 @@ export default class ProposalModel extends Model<ProposalAttributes> {
       return []
     }
 
+    let timeFrame = this.parseTimeframe(filter.timeFrame)
+
+    let orderBy = 'p.created_at'
```
andyesp, line 196: You can write this `const orderBy = filter.search || 'rank'`

src/entities/Proposal/routes.ts
```diff
@@ -128,24 +128,35 @@ function dropSnapshotProposal(proposal_space: string, proposal_id: string) {
...
+    Number(query.offset) : MIN_PROPOSAL_OFFSET
 
-  let limit = req.query.limit && Number.isFinite(Number(req.query.limit)) ?
-    Number(req.query.limit) : MAX_PROPOSAL_LIMIT
+  let limit = query.limit && Number.isFinite(Number(query.limit)) ?
+    Number(query.limit) : MAX_PROPOSAL_LIMIT
+
+
+  if (search) {
+    if (!/\w{3}/.test(search)) {
```
andyesp, line 152: You can write `if (search && !/\w{3}/.test(search)) {` to avoid a nested if.

src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  navigate(locations.proposals(newParams))
+}
+
+export default function SearchInput(props: React.HTMLAttributes<HTMLDivElement>) {
+  const l = useFormatMessage()
+  const location = useLocation()
+  const params = useMemo(() => new URLSearchParams(location.search), [location.search])
+  const search = useMemo(() => params.get('search') || '', [params])
```
andyesp, line 29: I don't think it's necessary to memo params and search as there is no expensive value to calculate that will change on every render.

1emu: Ty! I'm removing the one for search, here and in other components.
Are you sure about the params with the `new URLSearchParams`? I think by memoizing we could avoid creating a new instance on every render

andyesp: You're right!

1emu: @2fd do you have any input on this? we'd appreciate it! 😄

src/pages/index.tsx
```diff
@@ -68,9 +53,7 @@ export default function IndexPage() {
   const responsive = useResponsive()
   const isMobile = responsive({ maxWidth: Responsive.onlyMobile.maxWidth })
   const burgerMenu = useBurgerMenu();
-  const translate = '550px'
-
-  // const [ ff ] = useFeatureFlagContext()
+  const [burgerOpen, burgerTranslate] = useMemo(() => [burgerMenu?.status.open, burgerMenu?.status.translate], [burgerMenu?.status])
```
andyesp, line 56: Is this memo necessary? Are these causing multiple re-renders without memo?

src/components/Layout/BurgerMenu.tsx
```diff
@@ -1,22 +1,42 @@
...
+      menuContext?.status.open,
+      menuContext?.status.searching,
+      menuContext?.status.filtering]
+  }, [menuContext?.status])
 
-  const toggleMenuHandler = () => {
-    statusContext?.setStatus(prev => !prev)
+  const handleClick = () => {
+    if (!searching) {
+      menuContext?.setStatus(({ ...menuContext.status,
```
andyesp, line 18: As setStatus is a "setState" from useState, you can use here the previous state from setState instead of using the state. It may cause an issue if status is updated and this function didn't catch up with this update (it's not wrapped in a useCallback).
E.g.: `menuContext.setStatus(prevState => ({ ...prevState, open, filtering })`

src/components/Layout/BurgerMenuContent.tsx
```diff
@@ -1,42 +1,112 @@
...
+const STATUS_FILTER_HEIGHT = 290
+const TIMEFRAME_FILTER_HEIGHT = 212
+const CLOSED_FILTER_HEIGHT = 56
+const MOBILE_NAVIGATION_HEIGHT = 104
+
+function BurgerMenuContent({ navigationOnly }: BurgerMenuContentProps) {
   const [footer, setFooter] = useState<Element | null>(null)
   const burgerMenu = useContext(BurgerMenuStatusContext)
+  const [filterStatus, setFilterStatus] = useState(filtersInitialStatus);
+  const [burgerOpen, burgerSearching, burgerFiltering, burgerTranslate] = useMemo(() => {
```
andyesp, line 35: Are these triggering re-renders? I'd use `burgerMenu.status` directly instead of memoizing it.

src/hooks/useBurgerMenu.ts
```diff
@@ -1,19 +1,27 @@
...
   useEffect(() => {
-    burgerShow?.setShow(true)
-  
-    return () => {
-      burgerShow?.setShow(false)
-      burgerMenu?.setStatus(false)
-    };
-  }, []);
+      const filtering =  !!(type || status || timeFrame && timeFrame.length > 0)
+      burgerMenu?.setStatus({
```
andyesp, line 11: I'd use here `burgerMenu.setStatus(prevState => ({ ...prevState, ...rest })))`, also burgerMenu.status is not part of the dependencies of this effect, this can cause this effect to be outdated.

src/components/Search/SearchTitle.css
```diff
@@ -0,0 +1,61 @@
...
+
+.SearchTitle__Separator {
+  width: 100%;
+  border-color: #e1e0e3;
+  opacity: 0.2;
+  margin-top: 16px;
+}
+
+@media (max-width: 767px) {
+  .SearchTitle.OnlyMobile {
```
andyesp, line 42: Is `OnlyMobile` and `OnlyDesktop` classes necessary? You can handle this with media queries. Using this media query is "only mobile". I'd use a "mobile first" approach to avoid using max-width.

src/components/Search/SearchTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const responsive = useResponsive()
+  const { search } = useSearchParams()
+  const isMobile = responsive({ maxWidth: Responsive.onlyMobile.maxWidth })
+  const isTablet = responsive({ maxWidth: Responsive.onlyTablet.maxWidth })
+  const [size, maxLength] = useMemo(() => {
+    if (isMobile) return ['medium', 15]
+    if (isTablet) return ['large', 24]
+    else return ['huge', 23]
+  }, [isMobile, isTablet])
+  const formattedSearch = useMemo(() => search && search.length > maxLength ? search.substring(0, maxLength) : search, [search, maxLength])
```
andyesp, line 24: I think we can limit this with css using ellipsis. And the `size` variable with media queries.

src/hooks/useBurgerMenu.ts
1emu: 👌 with `burgerMenu.setStatus(prevState => ({ ...prevState, ...rest })))`!

The status is nor part of the dependencies because this effect is meant to calculate the initial status for the burger menu based on the search params (if any!), but it should run once :)

1emu:
Last version updated here: http://pr186.test.decentraland.vote/

src/components/Category/CategoryOption.tsx
```diff
@@ -33,7 +30,6 @@ export default React.memo(function CategoryOption({ active, type, className, ...
 
       if (props.href) {
         navigate(props.href)
-        burgerMenu?.setStatus(false)
```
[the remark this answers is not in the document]
1emu: Thx! Added this back again but from the `CategoryList.tsx` component.

yemel (approved):
LGTM

1emu:
Update:
@ncomerci found an issue with Firefox focus on inputs

image

Fixed it by removing that outline, since we already have a focus indicator 

image

src/components/Layout/BurgerMenuContent.tsx
```diff
@@ -1,42 +1,124 @@
...
+const STATUS_FILTER_HEIGHT = 290
+const TIMEFRAME_FILTER_HEIGHT = 212
+const CLOSED_FILTER_HEIGHT = 56
+const MOBILE_NAVIGATION_HEIGHT = 104
+
+function BurgerMenuContent({ navigationOnly }: BurgerMenuContentProps) {
   const [footer, setFooter] = useState<Element | null>(null)
   const burgerMenu = useContext(BurgerMenuStatusContext)
+  const [filterStatus, setFilterStatus] = useState(filtersInitialStatus)
+  const [burgerOpen, burgerSearching, burgerFiltering, burgerTranslate] = [
```
andyesp, line 35: You can do destructuring here `burgerMenu?.status` to avoid creating a new array
`const { burgerOpen, burgerSearching, burgerFiltering, burgerTranslate } = burgerMenu?.status`

src/components/Layout/BurgerMenu.tsx
```diff
@@ -1,24 +1,55 @@
...
+import React, { useContext, useMemo } from 'react'
 import { BurgerMenuStatusContext } from '../Context/BurgerMenuStatusContext'
 import TokenList from 'decentraland-gatsby/dist/utils/dom/TokenList'
 import './BurgerMenu.css'
 
 function BurgerMenu() {
-  const statusContext = useContext(BurgerMenuStatusContext)
-  const showContext = useContext(BurgerMenuShowContext)
+  const menuContext = useContext(BurgerMenuStatusContext)
+  const [show, open, searching, filtering] = [
```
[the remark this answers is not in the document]
2fd: +1

src/components/Layout/BurgerMenuContent.tsx
2fd: +1

2fd (changes requested):

andyesp (approved):

PR #191, opened 2022-02-23
Attempt to improve proposal filter wording

maraoz:
I found these categories or states a little confusing:
image

I tried to learn what each of those meant by filtering current proposals and doing some guesswork. 
What do you think of my wording changes? Not looking to merge this yet, just want to start a discussion.

Additionally, do you think maybe a "how does this work?" section could be added to the app, with a flow chart showing different states a proposal can be in, and how they move from one to another?

### DIFF
--- a/src/intl/en.json
+++ b/src/intl/en.json
@@ -173,13 +173,13 @@
     "governance_description": "Formalize the passed version of a Draft Proposal into a binding governance outcome"
   },
   "status": {
-    "all": "All outcomes",
-    "pending": "Pending outcomes",
-    "finished": "Finished outcomes",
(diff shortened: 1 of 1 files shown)


jparnaudo:
I like the new names and the idea of making a flow chart! 

We should also review if [this](https://github.com/decentraland/governance/#about-proposals-statuses) is still valid for all the proposal statuses

ginocingolani:
@maraoz We were doing grooming and realized this was still here. Thanks for the suggestion but we will take a more holistic approach, we need to redefine how proposal status works across the entire flow from creation to enaction, and then change the wording. It is on our radar so stay tuned!

Issue #199, opened 2022-03-17
Show all comments from the forum in the proposal page


ginocingolani:
At the moment we're only showing the latest comments sorted by date from the forum on the Governance dApp. Ref PR: https://github.com/decentraland/governance/pull/159
It would be ideal that once the user clicks on the Read More button, more comments are fetched and displayed on-page.

1emu:
When user presses 'read more':
- we show all user comments from the forum (this could be a few or several). **
- the 'Read More' button changes to 'Join the Discussion' and links to the forum

Would this be OK?

**A note on performance: The API request always returns the full topic with all the comments, so fetching a few or all of them doesn't make a different (we always get all)

ginocingolani:
If performance is not an issue I agree with showing all the comments at once and changing the button to Join the Discussion

PR #215, opened 2022-03-25
199 - Add button to show all comments on proposal

1emu:
This PR closes [this issue](https://github.com/decentraland/governance/issues/199)
Ready to be tested at: http://pr199.test.decentraland.vote/

When comments are equal or less than 3, it reads "Join the discussion"
image

When there are more than 3 comments available for display, it changes to "Read More"
image

With several comments
image

when opened:
image

Made some tweaks to mobile font sizes:
image

when opened:
image

### DIFF
--- a/src/components/Proposal/ProposalComments.css
+++ b/src/components/Proposal/ProposalComments.css
@@ -35,3 +35,12 @@
   border: 1px solid rgba(115,110,125,.32);
   width: 100%;
 }
+
+@media only screen and (max-width: 425px) {
+  .ProposalComments .ui.button {
+    font-size: 14px;
+  }
+  .ProposalComments .ui.header {
+    font-size: 17px;
+  }
+}
--- a/src/components/Proposal/ProposalComments.tsx
+++ b/src/components/Proposal/ProposalComments.tsx
@@ -117,7 +117,7 @@ export default React.memo(function ProposalComments({ proposal, loading, ...prop
                 rel="noopener noreferrer"
                 href={(proposal && forumUrl(proposal)) || ''}
               >
-                {t('page.proposal_comments.join_discussion_label')}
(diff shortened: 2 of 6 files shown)


ginocingolani:
Looks amazing! A couple of comments:
- Since loading performance is not an issue maybe we could increase the limit from 3 to 5 on the first load.
- To avoid visual redundancy (Seeing 'Join the conversation' twice) on the use case of 3 comments of less, we could update the copy of the button to Comment on this Proposal

src/components/Proposal/ProposalComments.css
```diff
@@ -35,3 +35,12 @@
   border: 1px solid rgba(115,110,125,.32);
   width: 100%;
 }
+
+@media only screen and (max-width: 425px) {
```
andyesp, line 39: Why `425px`?  I'd change this to min-width with a mobile first approach.
(We should talk to @heyimjuani about what breakpoints are we supporting.)

heyimjuani: Don't have this info off the top of my head but I'd be a +1 for going through the entire platform to review and normalize.

1emu: Hmm I'm not sure what approach is best.

I'm following what we did with @heyimjuani for the `Finish at` [label on the proposal list](https://github.com/decentraland/governance/blob/4951c4f8d08775bc89e9150ba2df956d97e1ddb8/src/components/Proposal/ProposalItem.css#L16).
The `425px` in this case is because from that screen size onwards, the font size doesn't need a fix to fit.

Do you think maybe we can use this pattern until we meet and review everything?

1emu:
Changes after @ginocingolani suggestions:

Bottom button shows `Comment on this Proposal`:
image

Initially load 5 comments:
image

image

image

Issue #137, opened 2021-11-10
fix: delegation info on balance page


2fd:
Example: https://governance.decentraland.org/balance/?address=0xb0145Ae156D201d6E371d07265FE3C045071c967

![governance decentraland org_balance__address=0xb0145Ae156D201d6E371d07265FE3C045071c967](https://user-images.githubusercontent.com/208789/141201775-f6170848-932b-4b0a-8dc3-249eebf6f5f3.png)

This user got is VP from a delegation from `0x539d5251360887da47e65c1dd65465ff581a6c5d`

Check the delegation query

PR #216, opened 2022-03-28
Fix: delegation info on balance page

ncomerci:
This PR addresses #137 

Spaces considered:

- "" (empty)
- "snapshot.dcl.eth"

### DIFF
--- a/src/hooks/useDelegation.ts
+++ b/src/hooks/useDelegation.ts
@@ -4,13 +4,13 @@ import useAsyncMemo from "decentraland-gatsby/dist/hooks/useAsyncMemo";
 const ENDPOINT = `https://api.thegraph.com/subgraphs/name/snapshot-labs/snapshot`
 const QUERY = `
 query ($space: String!, $address: String!) {
-  delegatedTo: delegations(where: { space: $space, delegator: $address }, first: 1, orderBy: timestamp, orderDirection: desc) {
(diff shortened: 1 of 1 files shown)


src/hooks/useDelegation.ts
```diff
@@ -4,13 +4,13 @@ import useAsyncMemo from "decentraland-gatsby/dist/hooks/useAsyncMemo";
 const ENDPOINT = `https://api.thegraph.com/subgraphs/name/snapshot-labs/snapshot`
 const QUERY = `
 query ($space: String!, $address: String!) {
-  delegatedTo: delegations(where: { space: $space, delegator: $address }, first: 1, orderBy: timestamp, orderDirection: desc) {
+  delegatedTo: delegations(where: { space_in: ["", $space], delegator: $address }, first: 1, orderBy: timestamp, orderDirection: desc) {
```
2fd, line 7: What happen if we get both delegations? shouldn't we handle the case where we get 2 responses?

2fd (changes requested):

2fd (approved):

src/hooks/useDelegation.ts
2fd: looks good to me

andyesp (approved):

andyesp (approved):

PR #217, opened 2022-03-28
Search improvements & minor fixes

1emu:
This PR can be tested [here](http://pr217.test.decentraland.vote/)

- Removed autofocus to the search input on page load
image

- Changed displayed search title

Example for short text
image

Example for long text
image

In mobile:
image

- Fix to sorting menu in mobile view (tx to @ncomerci )
image

### DIFF
--- a/src/components/Search/SearchTitle.css
+++ b/src/components/Search/SearchTitle.css
@@ -8,6 +8,10 @@
 	align-items: center;
 }
 
+.SearchTitle .SearchTitle_TextContainer {
+	display: contents;
+}
+
 .SearchTitle .dcl.back {
 	background-position-x: 9px;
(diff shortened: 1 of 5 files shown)


src/components/Search/SearchTitle.css
```diff
@@ -8,6 +8,10 @@
 	align-items: center;
 }
 
+.SearchTitle .SearchTitle_TextContainer {
+	display: contents;
```
andyesp, line 12: I'm not sure about this property, I thought this is for using in a grid context 🤔 but I couldn't find an issue with it.
If using `display: inline-flex` and `overflow: hidden` don't do the trick, then this is fine with me.

PR #218, opened 2022-03-30
Grant size validation on the fly

ncomerci:
In the `Request a grant` form, the behavior of the `Grant Size` field is now similar to that of the `Beneficiary address` field. As soon as you start typing, it validates the constrains and displays feedback to the user without the user having to press the Submit proposal button.


1emu (approved):
Tested locally, LGTM 👍

PR #252, opened 2022-04-11
Feat: Delegation Banner

### LINKED ISSUE #225: "Call for delegates open" banner
image

Save in local storage if user dismissed banner. Check newsletter subscription banner.
Pending UI update with dismiss button on top right corner.

ncomerci:
This PR closes #225 

image


src/components/Banner/Delegation/DelegationBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <Button className="Banner__Button" primary size="small" href={process.env.GATSBY_DELEGATE_APPLICATION_URL || ''}>
+      {label}
+    </Button>
+  )
+}
+
+function DelegationBanner({ active, bannerState }: DelegationBannerProps) {
+  const t = useFormatMessage()
+  const { setShowBanner } = bannerState
```
1emu, line 27: I'm thinking that maybe it'd be better if the component can handle it's state internally rather than receiving a state. 

It could receive just the initial configuration (if needed), but it's state should live inside of it (or the Banner component)

src/hooks/useNewsletterSubscriptionModal.ts
```diff
@@ -0,0 +1,31 @@
+import { useState } from 'react'
+import { BannerState, ShowBanner } from '../components/Banner/Banner'
+
+function useNewsletterSubscriptionModal(bannerState: BannerState) {
```
1emu, line 4: I'm wondering if we really need this hook. 
I think the `finishSubscription` and `closeSubscriptionBanner` are closely related to the subscription banner, and could be defined in the component.

My rationale here is: what is this giving us that the component and its state can't do by themselves? 🤔

1emu (changes requested):
I like the idea of having a Banner, but I have some doubts about the implementation. Left some comments so we can discuss it :)

src/pages/index.tsx
```diff
@@ -112,7 +136,7 @@ export default function IndexPage() {
         className="OnlyMobile Animated"
         style={(isMobile && open && { transform: 'translateX(-200%)', height: 0 }) || {}}
       >
-        <SubscriptionBanner active={!type} />
+        {subscriptionBanner}
```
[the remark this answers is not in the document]
1emu: I believe it's stored it in a variable because he's using it in two places, right?

src/components/Banner/Delegation/DelegationBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  useEffect(() => {
+    const showSubsBanner = localStorage.getItem(HIDE_DELEGATE_BANNER_KEY) != 'true'
+    setShowBanner(showSubsBanner ? ShowBanner.YES : ShowBanner.NO)
+  }, [])
+
+  const bannerProps: BannerProps = {
```
andyesp, line 34: Pass props directly to component in render tree instead of adding them to a variable please 🙏

src/components/Icon/Delegate.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+export default React.memo(function Delegate(props: React.SVGProps<SVGSVGElement>) {
```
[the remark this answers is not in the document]
andyesp: Also the props aren't needed here.

src/components/Icon/Email.tsx
```diff
@@ -0,0 +1,12 @@
+import  React from 'react'
+
+export default React.memo(function Email(props: React.SVGProps<SVGSVGElement>) {
```
andyesp, line 3: We don't need to memo the icons. I'd also avoid to pass all props and instead limit what props we need for icons.

src/hooks/useNewsletterSubscriptionModal.ts
andyesp: I think the state can be handled internally.

src/components/Banner/Delegation/DelegationBanner.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    const showSubsBanner = localStorage.getItem(HIDE_DELEGATE_BANNER_KEY) != 'true'
+    setShowBanner(showSubsBanner)
+  }, [])
+
+  return (
+    <div className="DelegationBanner__Container">
+      <Banner
+        isVisible={showBanner && isVisible}
```
andyesp, line 27: Why don't 

```
const isVisible = localStorage.getItem(HIDE_DELEGATE_BANNER_KEY) != 'true'

...

<Banner isVisible={isVisible} />
```

src/components/Modal/NewsletterSubscriptionModal.tsx
```diff
@@ -34,7 +32,7 @@ export type NewsletterSubscriptionModalProps = Omit<ModalProps, 'children'> & {
 
 async function subscribe(email: string) {
   try {
-    await Decentraland.get().subscribe(email!)
+    // await Decentraland.get().subscribe(email!)
```
andyesp, line 35: Forgot to remove this comment 🙈

src/components/Banner/Delegation/DelegationBanner.tsx
ncomerci: Remember that the banner is also hidden depending on whether there is an active search or not.

In `index.tsx` the banner is called this way: 

```html
<div className="OnlyDesktop">
  <DelegationBanner isVisible={!searching} />
</div>
```

src/components/Modal/NewsletterSubscriptionModal.tsx
ncomerci: My bad 😳

src/hooks/useNewsletterSubscriptionModal.ts
ncomerci: Today @andyesp was helping me to improve the code and we left the hook for 2 reasons:

- We must share logic between the subscription modal and the subscription banner.
- To group all those functions and variables to avoid enlarging the `index.tsx` component.

src/components/Banner/Delegation/DelegationBanner.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type DelegationBannerProps = {
+  isVisible: boolean
+}
+
+const HIDE_DELEGATE_BANNER_KEY = 'org.decentraland.governance.delegate_banner.hide'
+
+function DelegationBanner({ isVisible }: DelegationBannerProps) {
+  const t = useFormatMessage()
+  const [showBanner, setShowBanner] = useState(false)
```
andyesp, line 17: I think showBanner can be just a variable without a useState.

`const showBanner = localStorage.getItem(HIDE_DELEGATE_BANNER_KEY) != 'true'`

src/components/Banner/Delegation/DelegationBanner.tsx
andyesp: You are right. What I wanted to say is that you can remove the use effect and just store the localStorage item check in a variable and use that to pass to the isVisible prop.

`isVisible={isVisible && !hasHideDelegateBannerKey}`

src/hooks/useNewsletterSubscription.ts
```diff
@@ -0,0 +1,46 @@
...
+    setSubscribed(false)
+  }
+
+  return {
+    showSubscriptionBanner: showBanner,
+    isSubscriptionModalOpen: isOpen,
+    setIsSubscriptionModalOpen: setIsOpen,
+    onSubscriptionSuccess: finishSubscription,
+    subscribed,
+    onclose: closeSubscriptionBanner
```
1emu, line 41: onClose :)

src/pages/index.tsx
```diff
@@ -63,6 +68,15 @@ export default function IndexPage() {
   const { status: burgerStatus } = useBurgerMenu()
   const { open, translate } = burgerStatus
 
+  const {
+    showSubscriptionBanner,
+    isSubscriptionModalOpen,
+    setIsSubscriptionModalOpen,
+    onSubscriptionSuccess,
+    subscribed,
+    onclose,
```
1emu, line 77: onClose inThisFileAlsoPlease

1emu (approved):
Looks great! Ty @ncomerci @andyesp  🥁 🎸 
Left a tiny comment for a missing  🐫 case

src/hooks/useNewsletterSubscriptionModal.ts
1emu: Beautiful ✨ 🦋

src/components/Banner/Delegation/DelegationBanner.tsx
ncomerci: localStorage must be inside a useEffect
image

PR #266, opened 2022-04-18
Delegate in proposal voting section

andyesp:


src/components/Section/ChoiceButton.tsx
```diff
@@ -9,9 +9,10 @@ export type ChoiceButtonProps = Omit<ButtonProps, 'children' | 'color'> & {
   choice?: string
   voted?: boolean
   color?: ChoiceColor
+  text: string
 }
 
-export default function ChoiceButton({ voted, color, choice, ...props }: ChoiceButtonProps) {
+export default function ChoiceButton({ voted, color, choice, text, ...props }: ChoiceButtonProps) {
```
1emu, line 15: We are no longer using choice, right?

1emu (approved):

src/components/Section/DelegateLabel.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    if (vote && vote?.choice !== delegateVote.choice) {
+      return t('page.proposal_detail.delegate_voted_differently')
+    }
+  }
+
+  return ''
+}
+
+interface Props {
```
ncomerci, line 69: Shouldn't this be named DelegateLabelProps ?

src/components/Section/ProposalVotingSection.tsx
```diff
@@ -0,0 +1,129 @@
...
+import useAuthContext from 'decentraland-gatsby/dist/context/Auth/useAuthContext'
+import { Vote } from '../../entities/Votes/types'
+import locations from '../../modules/locations'
+import { calculateChoiceColor } from '../../entities/Votes/utils'
+import ChoiceButton from './ChoiceButton'
+import useDelegation from '../../hooks/useDelegation'
+import DelegateLabel from './DelegateLabel'
+import VotedChoiceButton from './VotedChoiceButton'
+
+interface Props {
```
ncomerci, line 16: Same

src/components/Section/VotedChoiceButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  return {
+    color: 1,
+    text: '',
+  }
+}
+
+interface Props {
```
ncomerci, line 44: Same

src/components/Section/DelegateLabel.tsx
1emu: Hi @ncomerci ! 

I mentioned the same to @andyesp, but we agreed that since we are not exporting the Props interface or using it anywhere else, it's ok to name it `Props`.
If that weren't the case, then yeah, it would be better to call it with the full name of the component it get's used on, so it's clearer outside that file.

What do you think?
Same would apply for the other comments.

ncomerci: Sounds good to me!

PR #267, opened 2022-04-20
Feat: VP Delegation Modal

### LINKED ISSUE #229: Voting power delegation modal
Screen Shot 2022-04-04 at 15 16 29

“Pick somebody else” would take the user to [Snapshot](https://snapshot.org/#/delegate/snapshot.dcl.eth)[https://snapshot.org/#/delegate/snapshot.dcl.eth](https://www.google.com/url?q=https://snapshot.org/%23/delegate/snapshot.dcl.eth&sa=D&source=docs&ust=1649093959757596&usg=AOvVaw1MWFDNPr_XUXNOsoOACpEb) so they can enter another user to delegate their VP.

“picked by” data is the amount of users who chose the delegate. (Snapshot)

List should display Delegated VP (not total).

Candidate list from JSON in the repository.

Clicking candidate shows Candidate Detail (#230).

Pending:
* Read more URL @ginocingolani

ncomerci:
This PR closes #229 

Staging server: http://pr267.test.decentraland.vote/balance/

### Missing
- "Read more" link
- Picked by

### Screenshot
screencapture-localhost-8000-balance-2022-04-20-12_58_46


src/components/Modal/VPDelegationModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  return address.substring(0, 6) + '...' + address.substring(38, 42)
+}
+
+function tableCell(delegate: Delegate, picked_by: number, intl: IntlShape, arrowRef: React.RefObject<HTMLDivElement>) {
+  return (
+    <>
+      <Table.Cell>
+        <span className="Candidate">
+          <Avatar address={delegate.address} size="small" />
+          {delegate.ens_name || addressShortener(delegate.address)}
```
1emu, line 41: There is an `Address` component that already shortens address for us!

With @andyesp we were also thinking of creating a `Username` component that we can reuse, because we are calling it in several places (like [here](https://github.com/andyesp/governance/blob/46c9684c07c8dfb14f3ddea4f8e9a60fc00b0667/src/components/Section/DelegateLabel.tsx#L13))

```typescript
const Username = ({ address }: { address: string }) => {
  const { profile } = useProfile(address)

  return (
    <Link className="DetailsSection__Value" href={locations.balance({ address: address })}>
      {profile && profile.name && <Avatar size="mini" address={profile.ethAddress} style={{ marginRight: '.5rem' }} />}
      {profile && profile.name}
      {(!profile || !profile.name) && !!address && (
        <Blockie scale={3} seed={address || ''}>
          <Address value={address || ''} />
        </Blockie>
      )}
    </Link>
  )
}
```

I'm thinking that you might not even need the ens name in the json because the `useProfile` hook already looks it up for you with the address 🤔 

Maybe we could take this into account and push the refactored component to the delegation branch, and make it so it can receive some classes/styles as props, so we all can use it? What do you say @andyesp?

1emu (commented):
The modal looks sick 🔥 !
I left some refactoring suggestions I'd like for us to consider.

src/components/Modal/VPDelegationModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+                  onMouseEnter={() => arrowRefs[idx].current?.classList.add('Filled')}
+                  onMouseLeave={() => arrowRefs[idx].current?.classList.remove('Filled')}
+                  onClick={() => {
+                    navigate(locations.balance({ address: dgt.address }))
+                    setTimeout(() => {
+                      props.onClose()
+                    }, 100)
+                  }}
+                >
+                  {tableCell(dgt, 0, intl, ref)}
```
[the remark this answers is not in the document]
andyesp: Instead of using a function to render a component, I'd move this to a react component, e.g: `<VotingPowerDelegationItem />`

src/components/Modal/VPDelegationModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <Table.Cell>{intl.formatNumber(delegate.total_vp)}</Table.Cell>
+      <Table.Cell>
+        <div ref={arrowRef} className="Arrow" />
+      </Table.Cell>
+    </>
+  )
+}
+
+function VPDelegationModal({ vp, space, ...props }: VPDelegationModalProps) {
+  const EDIT_DELEGATION_URL = snapshotUrl(`#/delegate/${space}`)
```
andyesp, line 54: I'd rename this to `editDelegationUrl` as it is not a constant.

src/components/Modal/VPDelegationModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <Table.Cell>{intl.formatNumber(picked_by)}</Table.Cell>
+      <Table.Cell>{intl.formatNumber(delegate.total_vp)}</Table.Cell>
+      <Table.Cell>
+        <div ref={arrowRef} className="Arrow" />
+      </Table.Cell>
+    </>
+  )
+}
+
+function VPDelegationModal({ vp, space, ...props }: VPDelegationModalProps) {
```
andyesp, line 53: As @1emu said with `dgt`, I'd also avoid it here with `VP` and write `VotingPowerDelegationModal`

src/components/Modal/VPDelegationModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const EDIT_DELEGATION_URL = snapshotUrl(`#/delegate/${space}`)
+
+  const t = useFormatMessage()
+  const intl = useIntl()
+  const [isDescending, setIsDescending] = useState(true)
+
+  const delegates_with_total_vp = exampleDelegates.map((dgt) => {
```
andyesp, line 60: Use camelCase for variables here 🙏  (I think we only use snake case on backend model attributes)

ncomerci: My bad! This variable will be deleted when "picked_by" is resolved.

src/components/Modal/VPDelegationModal.tsx
ncomerci: - addressShortener replaced by the Address component
- Use of the hook useProfile to get the name

Maybe the 3 of us could discuss how to refactor this component because it is not clear to me how to make it generic for the different use cases.

src/components/Icon/Arrow.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+
+type ArrowProps = {
+  filled: boolean
+}
+
+function Arrow({ filled }: ArrowProps) {
+  const colors = {
+    fill: 'white',
```
andyesp, line 9: You can write this:
```
const fill = filled ? '#736E7D' : 'white'
const stroke = filled ? 'white' : '#736E7D'
```
and avoid creating an object.

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationModal.css
```diff
@@ -0,0 +1,67 @@
...
+  font-weight: 600 !important;
+  margin-bottom: 10px !important;
+}
+
+.VotingPowerDelegationModal .description .dg.Paragraph {
+  font-size: 15px;
+}
+
+.VotingPowerDelegationModal .description .dg.Paragraph .dg.Link {
+  color: var(--blue);
```
1emu, line 12: ✨

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationModal.tsx
```diff
@@ -0,0 +1,104 @@
...
+              <Table.HeaderCell className="TotalVP" onClick={() => setIsDescending((prev) => !prev)}>
+                <span>
+                  {t('modal.vp_delegation.total_vp')}
+                  <Sort rotate={isDescending ? 0 : 180} />
+                </span>
+              </Table.HeaderCell>
+              <Table.HeaderCell />
+            </Table.Row>
+          </Table.Header>
+          <Table.Body>{delegates.map(createDelegateRow)}</Table.Body>
```
1emu: 😍

src/components/Subscription/SubscriptionBanner.css
```diff
@@ -16,7 +16,7 @@
   position: absolute;
   top: 0;
   left: 0;
-  background-color: #4888EE;
+  background-color: var(--blue);
```
1emu, line 19: ty

src/hooks/useDelegatesInfo.ts
```diff
@@ -0,0 +1,107 @@
...
+        method: 'post',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          query: QUERY,
+          variables: { address: addresses, space: SNAPSHOT_SPACE },
+        }),
+      }
+    )
+
+    const body = await request.json()
```
1emu, line 64: I'm wondering, what happens if the fetch fails? 🤔

src/hooks/useDelegatesInfo.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      const delegations = queryResult.filter(deleg => deleg.delegate === addr)
+      pickedBy.set(addr, new Set())
+
+      if (delegations.length > 0) {
+        for (const deleg of delegations) {
+          pickedBy.get(deleg.delegate)?.add(deleg.delegator)
+        }
+      }
+    }
```
1emu, line 79: Maybe the `picked by` could be calculated by filtering all delegations where the delegate has the address you are querying for, and getting the length?

```typescript
const result = PickedByResult[]
for (const address of addresses) {
    const delegations = queryResult.filter(delegation => delegation.delegate === address)
    result.push({ address, delegations.length })
 }
```

src/hooks/useDelegatesInfo.ts
```diff
@@ -0,0 +1,107 @@
...
+function useDelegatesInfo(addresses: string[]): Delegate[] {
+
+  addresses = addresses.map(addr => addr.toLowerCase())
+
+  const totalVotingPower = getTotalVP(addresses)
+  const [pickedByResults] = getPickedBy(addresses)
+
+  return useMemo(
+    () => pickedByResults.map(delegate => ({ ...delegate, totalVP: totalVotingPower[delegate.address] })),
+    [JSON.stringify(pickedByResults), JSON.stringify(totalVotingPower)]
```
1emu, line 103: Is stringifying the memo dependencies more efficient? 😮

src/hooks/useDelegation.ts
```diff
@@ -80,7 +81,7 @@ export default function useDelegation(address?: string | null) {
     }
 
     const request = await fetch(
-      ENDPOINT,
+      SNAPSHOT_QUERY_ENDPOINT,
```
1emu, line 84: 😄

1emu (approved):
LGTM! Left some questions/suggestions

andyesp (approved):

src/hooks/useDelegatesInfo.ts
ncomerci: In that case you would be double counting someone who delegated in the dcl space and in the global space to the same person

src/hooks/useDelegatesInfo.ts
ncomerci: I was having strange behavior without stringify and this solved it! https://stackoverflow.com/questions/59467758/passing-array-to-useeffect-dependency-list

src/hooks/useDelegatesInfo.ts
ncomerci: try-catch implemented!

PR #274, opened 2022-04-27
Add voting power stats to delegates detail modal

andyesp:
image

image

### DIFF
--- a/src/api/Snapshot.ts
+++ b/src/api/Snapshot.ts
@@ -8,26 +8,26 @@ export type SnapshotQueryResponse<T> = { data: T }
 export type SnapshotResult = { ipfsHash: string }
 
 export type SnapshotStatus = {
-  name: string,
-  network: string,
-  version: string,
-  tag: string,
-  relayer: string,
+  name: string
+  network: string
+  version: string
+  tag: string
+  relayer: string
 }
 
 export type SnapshotStrategy = {
   name: string
   params: {
-    symbol: string,
-    address: string,
+    symbol: string
+    address: string
     decimals: number
   }
 }
 
 export type SnapshotSpace = {
-  id: string,
-  network: string,
-  strategies: SnapshotStrategy[],
+  id: string
+  network: string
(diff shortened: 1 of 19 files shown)


src/api/Snapshot.ts
```diff
@@ -212,16 +205,56 @@ export class Snapshot extends API {
...
+          orderDirection: asc
+        ) {
+          id
+          voter
+          created
+        }
+      }
+    `
+
+    let votes: SnapshotVote[] = []
```
1emu, line 249: all this part is the same as the one in `getProposalVotes`, but instead of sending the proposal we are using the address. Maybe we can extract this to a fn?

src/components/Modal/VotingPowerDelegationDetail/VotingPowerDelegationDetail.tsx
```diff
@@ -12,14 +31,96 @@ type VotingPowerDelegationDetailProps = {
 
 function VotingPowerDelegationDetail({ delegate, onBackClick }: VotingPowerDelegationDetailProps) {
   const t = useFormatMessage()
+  const { address } = delegate
+  const { votingPower } = useVotingPowerBalance(address)
```
1emu, line 35: This reminded me of the `isLoadingVotingPower`... 
Maybe we could have a loader show if any of this information is missing?
(I think most of these methods already come with a `state` we can use)

1emu (approved):

src/api/Snapshot.ts
andyesp: Agree! Maybe we can talk later and see how can we extract it.

src/components/Modal/VotingPowerDelegationDetail/VotingPowerDelegationDetail.tsx
andyesp: Right! Let's handle all the loading cases after joining all the pieces of this modal

PR #279, opened 2022-05-02
Show list of voted proposals in candidate detail

andyesp:
Screen Shot 2022-05-02 at 17 01 32
image

Missing type of proposal and status. Figuring out best way to fetch all proposals from our db as this data is not in snapshot.


1emu (approved):
⭐

PR #291, opened 2022-05-09
Added useVotingPowerBalanceList hook

ncomerci:
useVotingPowerBalanceList receives an array of addresses and returns the VP of each address

### DIFF
--- a/src/api/Snapshot.ts
+++ b/src/api/Snapshot.ts
@@ -309,15 +309,13 @@ export class Snapshot extends API {
       block
     )
 
+    for (const addr of addresses) {
+      result[addr] = 0
+    }
+
     for (const score of scores) {
       for (const address of Object.keys(score)) {
         result[address] = (result[address] || 0) + Math.floor(score[address] || 0)
-      }
-    }
-
-    for (const addr of addresses) {
-      if (!result[addr]) {
-        result[addr] = 0
       }
     }
 


src/api/Snapshot.ts
```diff
@@ -310,9 +311,17 @@ export class Snapshot extends API {
 
     for (const score of scores) {
       for (const address of Object.keys(score)) {
-        result[address.toLowerCase()] = (result[address.toLowerCase()] || 0) + Math.floor(score[address] || 0)
+        const addr = address.toLowerCase()
```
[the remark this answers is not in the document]
ncomerci: Yes sorry, I forgot to change that line. thanks!

src/components/Section/ProposalVoting/VoteVotingPowerInfo.tsx
```diff
@@ -10,13 +10,14 @@ import './DelegationsLabel.css'
...
-  vote: Vote | null
-  account: string | null
+  vote?: Vote | null
+  account?: string | null
 }
 
 const VoteVotingPowerInfo = ({ vote, account }: VotingSectionFooterProps) => {
-  const { votingPower, isLoadingVotingPower } = useVotingPowerBalance(account)
-  const hasEnoughVP = !!votingPower && votingPower > 0 && !isLoadingVotingPower
+  account = account?.toLowerCase()
```
andyesp, line 18: I'd avoid re-assigning props.

1emu (approved):

1emu:
I like the idea of having just one method that knows how to execute the query depending on the amount of addresses received, but maybe we can call also have a hook that makes it easier to use if it's only one address:

```typescript
function useVotingPowerBalance(userAddress:string){
  const { votingPower, isLoadingVotingPower, hasEnoughToVote } = useVotingPowerBalances([userAddress])
  
  return {
    votingPower[userAddress],
    isLoadingVotingPower: state.loading,
    hasEnoughToVote[userAddress],
  }
}
```

andyesp (approved):

src/api/Snapshot.ts
```diff
@@ -310,9 +311,16 @@ export class Snapshot extends API {
...
     for (const score of scores) {
       for (const address of Object.keys(score)) {
-        result[address.toLowerCase()] = (result[address.toLowerCase()] || 0) + Math.floor(score[address] || 0)
+        result[address] = (result[address] || 0) + Math.floor(score[address] || 0)
       }
     }
+
+    for (const addr of addresses) {
+      if (!result[addr]) {
+        result[addr] = 0
```
1emu, line 320: If the idea of this is to initialize results for those addresses that aren't in score, I believe you can just move this up

```typescript
for (const addr of addresses) {
  result[addr] = 0
}

for (const score of scores) {
...
```

1emu (approved):
LGTM!

PR #298, opened 2022-05-10
Feat: Delegate VP button

ncomerci:
Direct interaction with the snapshot delegation contract

image


src/hooks/useSnapshotDelegateContract.ts
```diff
@@ -0,0 +1,38 @@
...
+import useAuth from 'decentraland-gatsby/dist/hooks/useAuth'
+import { useMemo } from 'react'
+import { SNAPSHOT_SPACE } from '../entities/Snapshot/constants'
+import DelegateABI from '../modules/contracts/abi/Delegate.abi.json'
+
+function useSnapshotDelegateContract() {
+
+  const [userAddress, authState] = useAuth()
+
+  return useMemo(() => {
```
andyesp, line 13: Is this memo necessary? I think we can safely remove it.

src/hooks/useSnapshotDelegateContract.ts
```diff
@@ -0,0 +1,38 @@
...
+    let setDelegate = null
+    let clearDelegate = null
+    let checkDelegation = null
+
+    if (authState.provider) {
+      const enc = new TextEncoder()
+      const signer = new Web3Provider(authState.provider).getSigner()
+      const spaceId = hexlify(enc.encode(SNAPSHOT_SPACE))
+      const fullSpaceId = spaceId.concat(new Array(66 - spaceId.length + 1).join('0'))
+      const contractAddress = process.env.GATSBY_SNAPSHOT_DELEGATE_CONTRACT_ADDRESS || ''
```
andyesp, line 25: This variable can be moved outside of function.

src/hooks/useSnapshotDelegateContract.ts
```diff
@@ -0,0 +1,38 @@
...
+import { SNAPSHOT_SPACE } from '../entities/Snapshot/constants'
+import DelegateABI from '../modules/contracts/abi/Delegate.abi.json'
+
+function useSnapshotDelegateContract() {
+
+  const [userAddress, authState] = useAuth()
+
+  return useMemo(() => {
+
+    let isContractUsable = false
```
andyesp, line 15: Can you write this like `const isContractUsable = authState.provider`?

src/hooks/useSnapshotDelegateContract.ts
```diff
@@ -0,0 +1,38 @@
...
+    let setDelegate = null
+    let clearDelegate = null
+    let checkDelegation = null
+
+    if (authState.provider) {
+      const enc = new TextEncoder()
+      const signer = new Web3Provider(authState.provider).getSigner()
```
andyesp, line 22: I think you can write this `const signer = authState.provider ? new Web3Provider(authState.provider).getSigner() : undefined`

andyesp: And remove `authState.provider` if condition

src/components/Modal/VotingPowerDelegationDetail/VotingPowerDelegationDetail.tsx
```diff
@@ -60,8 +64,31 @@ function VotingPowerDelegationDetail({ userVotes, candidate, onBackClick }: Voti
   const [isExpanded, setIsExpanded] = useState(false)
   const [matchingVotes, setMatchingVotes] = useState<MatchResult | null>(null)
   const [showFadeout, setShowFadeout] = useState(true)
+  const [isDelegating, setIsDelegating] = useState(false)
   const [filteredCandidateVotes, setFilteredCandidateVotes] = useState<VotedProposal[]>([])
 
+  const [delegatedAddress] = useAsyncMemo(
```
andyesp, line 70: I think this hook can be inside useSnapshotDelegateContract and useSnapshotDelegateContract can return the delegatedAddress.

src/intl/en.json
```diff
@@ -323,7 +323,10 @@
       "candidate_name": "Candidate name",
       "picked_by": "Picked by",
       "total_vp": "Total vp",
-      "button": "Pick somebody else",
+      "delegate_button": "Delegate VP",
+      "revoke_delegation_button": "Revoke delegation",
+      "delegated_stats": "Delegated {VP} VP",
```
1emu, line 328: this variable could be in lowercase :)

src/hooks/useSnapshotDelegateContract.ts
1emu: I'm not sure of this approach because later you'd need to check if signer is undefined or not...

Maybe we can write it like 

```typescript
  let setDelegate = null
  let clearDelegate = null
  let checkDelegation = null
  
  if(authState.provider && contractAddress){
    // ...
   }
   
   const contractIsUsable = setDelegate && clearDelegate && checkDelegation
   
   return { contractIsUsable, setDelegate, clearDelegate, checkDelegation }
```

src/hooks/useSnapshotDelegateContract.ts
1emu: When `authState.provider` changes we should update all methods right?

1emu (commented):
Nice work!
Left some comments we can debate 😄

src/hooks/useSnapshotDelegateContract.ts
1emu: Maybe we could even avoid the `isContractUsable` variable, and just do
`return {setDelegate, clearDelegate, checkDelegation }` since we already need to ask each time we want to use one of the contract functions.

We can do it like 

`isContractUsable && clearDelegate! ` => `clearDelegate && clearDelegate()`
`isContractUsable && setDelegate!(address)` => `setDelegate && setDelegate(address)`

```typescript
if (isContractUsable) {
        return await checkDelegation!()
      }
      return null
    },
```
=> `return checkDelegation && await checkDelegation()` ?

src/hooks/useSnapshotDelegateContract.ts
```diff
@@ -0,0 +1,39 @@
...
+  const [userAddress, authState] = useAuth()
+
+  return useMemo(() => {
+
+    let isContractUsable = false
+    let setDelegate = null
+    let clearDelegate = null
+    let checkDelegation = null
+
+    const contractAddress = process.env.GATSBY_SNAPSHOT_DELEGATE_CONTRACT_ADDRESS
```
andyesp, line 20: I meant outside of function / hook 😅  as a constant.

src/hooks/useSnapshotDelegateContract.ts
andyesp: I wouldn't initialize the functions as null either... I'd rather check "if contract is usable" inside of each function. But always expect a function from the component using the hook.

src/hooks/useSnapshotDelegateContract.ts
ncomerci: I used the memo thinking about what Lemu says, I do not see how it could be safely removed

andyesp (approved):
🚀 🪝

PR #307, opened 2022-05-13
Voting with calculation

1emu:
Use VP at the proposal time of creation instead of the latest VP.

### DIFF
--- a/src/components/Section/ProposalVoting/ProposalVotingSection.tsx
+++ b/src/components/Section/ProposalVoting/ProposalVotingSection.tsx
@@ -5,9 +5,11 @@ import useFormatMessage from 'decentraland-gatsby/dist/hooks/useFormatMessage'
 import { Button } from 'decentraland-ui/dist/components/Button/Button'
 import { Loader } from 'decentraland-ui/dist/components/Loader/Loader'
 
+import { ProposalAttributes } from '../../../entities/Proposal/types'
 import { Vote } from '../../../entities/Votes/types'
+import useDelegation from '../../../hooks/useDelegation'
 import useVotesMatch from '../../../hooks/useVotesMatch'
-import useVotingPowerInformation from '../../../hooks/useVotingPowerInformation'
+import useVotingPowerOnProposal from '../../../hooks/useVotingPowerOnProposal'
 import useVotingSectionTestData, { TestData } from '../../../hooks/useVotingSectionTestData'
 import { getPartyVotes, getVotingSectionConfig } from '../../../modules/votes/utils'
 import VotingSectionTester from '../VotingSectionTester'
@@ -17,7 +19,10 @@ import DelegationsLabel from './DelegationsLabel'
 import VotedChoiceButton from './VotedChoiceButton'
 import VotingSectionFooter from './VotingSectionFooter'
 
+import './ProposalVotingSection.css'
+
 interface Props {
+  proposal?: ProposalAttributes | null
   votes?: Record<string, Vote> | null
   loading?: boolean
   changingVote?: boolean
@@ -29,6 +34,7 @@ interface Props {
 }
 
 const ProposalVotingSection = ({
+  proposal,
   votes,
   loading,
   changingVote,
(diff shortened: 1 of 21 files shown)


src/components/Section/ProposalVoting/ProposalVotingSection.tsx
```diff
@@ -68,56 +82,71 @@ const ProposalVotingSection = ({
...
+    // TODO remove test data
+    [testData ? testData : [votes, choices, delegate, delegators, account, ownVotingPower, delegatedVp, voteDifference]]
+  )
+  const { votesByChoices, totalVotes } = useMemo(
+    () => getPartyVotes(delegators, votes, choices),
+    [delegators, votes, choices]
   )
-  const { votesByChoices, totalVotes } = useMemo(() => getPartyVotes(delegators, votes, choices), [testData])
 
+  loading = loading || accountState.loading || delegationState.loading || vpOnProposalState.loading
```
andyesp, line 96: Please don't reassign props 🙏  Maybe loading from props should have a different name. What is it loading? The proposal?

src/entities/Votes/utils.ts
```diff
@@ -159,3 +162,22 @@ export function abbreviateNumber(vp: number) {
 
   return scaled.toFixed(1) + suffix
 }
+
+export async function getProposalScores(proposal: ProposalAttributes, addresses: string[]) {
+  const result = {} as Record<string, number>
+  for (const addressesChunck of chunk(addresses, 500)) {
```
andyesp, line 168: Typo, `addressesChunk` 😁

src/components/Section/ProposalVoting/ProposalVotingSection.tsx
```diff
@@ -68,56 +82,71 @@ const ProposalVotingSection = ({
...
-        hasDelegators={delegators && delegators.length > 0}
-        started={started}
-        finished={finished}
-        account={account}
-        changingVote={changingVote}
-        onChangeVote={onChangeVote}
-      />
+          {delegationsLabel && <DelegationsLabel {...delegationsLabel} />}
+
+          {(showChoiceButtons || changingVote) && (
```
andyesp, line 121: I know you didn't change this but... maybe the check of `changingVote` can be in `showChoiceButtons`?

src/entities/Votes/utils.ts
```diff
@@ -159,3 +162,22 @@ export function abbreviateNumber(vp: number) {
 
   return scaled.toFixed(1) + suffix
 }
+
+export async function getProposalScores(proposal: ProposalAttributes, addresses: string[]) {
+  const result = {} as Record<string, number>
```
andyesp, line 167: Maybe you can move `Record<string, number>` to a type and reuse it in this function.
`type Scores = Record<string, number>`

src/components/Section/ProposalVoting/ProposalVotingSection.tsx
1emu: loading is `voting || proposalState.loading || votesState.loading || isLoadingVotingPower`.

Maybe `const proposalVotingSectionLoading = loading || accountState.loading || delegationState.loading || vpOnProposalState.loading`

src/components/Section/ProposalVoting/ProposalVotingSection.tsx
1emu: The thing is `showChoiceButtons` comes from the getConfig fn, and if we want it to include `changingVote` we'd need to pass it as a parameter. It'd be better if we can avoid calling it each time `changingVote` changes.

Maybe `showChoiceButtons` should be called `showVotingOptions` ? 🤔

andyesp: Oh ok, maybe you can keep this name.

andyesp (approved):

src/components/Section/VotingSectionTester.css
```diff
@@ -1,17 +1,17 @@
...
 	width: 100%;
-	padding-bottom: 16px;
 	align-items: center;
 	flex-direction: column;
 }
 
 .VotingSectionTester .button.ui.basic.button {
 	border: 1px solid rgba(115, 110, 125, .32);
-	background: #00800017 !important;
+	background: #5c5c6e !important;
```
ncomerci, line 10: We are using this color in many parts of the code, maybe it would be better to set a variable for it

image

src/hooks/useVotingPowerOnProposal.ts
```diff
@@ -0,0 +1,60 @@
...
+import { ProposalAttributes } from '../entities/Proposal/types'
+import { Vote } from '../entities/Votes/types'
+import { Scores, getProposalScores } from '../entities/Votes/utils'
+
+export const MINIMUM_VP_REQUIRED_TO_VOTE = 1
+
+function getDelegatedVotingPowerOnProposal(
+  scoresAtProposalCreation: Scores,
+  delegators: string[] | null,
+  votes: Record<string, Vote> | null | undefined
```
ncomerci, line 12: This could be rewritten as `votes?: Record<string, Vote> | null` 

It is the same but I think it looks cleaner 😃

ncomerci (approved):
✨
Only two non-blocking comments!

PR #315, opened 2022-05-18
Add hook to get delegations on proposal's blocknumber

1emu:
We were using latest delegation info instead of delegations in proposal snapshot.

Now:

- Proposal created before delegation says there's no delegates for this proposal

image

- Delegations label for proposal created after delegation correctly displays delegate

image


src/hooks/useDelegation.ts
```diff
@@ -63,41 +63,56 @@ export function filterDelegationFrom(delegations: Delegation[], space: string):
...
-      {
-        method: 'post',
-        headers: { 'Content-Type': 'application/json' },
-        body: JSON.stringify({
-          query: QUERY,
-          variables: { address: address.toLowerCase(), space: SNAPSHOT_SPACE },
-        }),
-      }
-    )
+export async function fetchAndFilterDelegates(query: string, variables: any) {
```
andyesp, line 74: Maybe this function can be part of the Snapshot api file.

andyesp (approved):

andyesp (approved):
⚡️

PR #328, opened 2022-05-24
Feat: Vesting section on enacted grants

### LINKED ISSUE #327: Add a link to the vesting contract on the enacted grant page
image

ncomerci:
This PR closes #327 

image


src/components/Section/VestingSection.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+function VestingSection({ vestingAddress }: Props) {
+  const vestingUrl = process.env.GATSBY_VESTING_DASHBOARD_URL
+
+  if (!vestingUrl) {
+    console.error('Vesting Dashboard URL not found')
+    return <></>
+  }
+
+  const t = useFormatMessage()
```
andyesp, line 21: Avoid rendering a hook conditionally. Please place this before the vesting Url condition.

src/components/Section/VestingSection.css
```diff
@@ -0,0 +1,22 @@
+.VestingSection {
+  padding: 9px 11px;
```
andyesp, line 2: I think this component needs more vertical padding. Check Figma!

andyesp (changes requested):

src/components/Section/VestingSection.tsx
ncomerci: Thank you! In this particular case I think it would not have been a problem because the environment variable is either there or it is not, but just to be sure I placed the hook before the condition.

1emu (approved):
🦺

andyesp (approved):

PR #370, opened 2022-06-13
Error handling improvements

1emu:
Admin section to try http requests with sleep timers.

image


src/pages/admin.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { Field } from 'decentraland-ui/dist/components/Field/Field'
+import { Header } from 'decentraland-ui/dist/components/Header/Header'
+
+import { HttpStat } from '../api/HttpStat'
+import ContentLayout, { ContentSection } from '../components/Layout/ContentLayout'
+import Navigation, { NavigationTab } from '../components/Layout/Navigation'
+import LogIn from '../components/User/LogIn'
+import useIsAdmin from '../hooks/useIsAdmin'
+
+import './transparency.css'
```
andyesp, line 20: I think this import is not necessary :)

src/entities/Admin/isAdmin.ts
```diff
@@ -0,0 +1,19 @@
...
+export const ADMIN_ADDRESSES = (env('ADMIN_ADDRESSES', '') || '')
+  .split(',')
+  .filter(isEthereumAddress)
+  .map((address) => address.toLowerCase())
+
+const adminAddresses = new Set(ADMIN_ADDRESSES)
+
+adminAddresses.forEach((address) => console.log('admin address:', magenta(address)))
+
+export default function isAdmin(user?: string | null | undefined) {
```
andyesp, line 14: Is this function used? 
I'd move this two lines inside the function
```
const adminAddresses = new Set(ADMIN_ADDRESSES)

adminAddresses.forEach((address) => console.log('admin address:', magenta(address)))
```

1emu: Right, I removed the fn and left that code outside so it runs on start, and it logs the addresses like the committee ones :)

andyesp (approved):

PR #383, opened 2022-06-21
Error display component

1emu:
Component shows on error

image

image

Expands showing received error message

image

image

image


src/components/Error/ErrorMessage.css
```diff
@@ -0,0 +1,103 @@
...
+  vertical-align: middle;
+  padding: 0 6px;
+  font-weight: 400;
+  font-size: 11px;
+  line-height: 26px;
+  letter-spacing: -0.2px;
+  text-transform: uppercase;
+  color: #736e7d;
+  background: #fff;
+  border: 1px solid rgba(115, 110, 125, 0.5);
```
ncomerci, line 73: This color is already defined as `--dark-gray-rgb`, you can write it as `rgba(var(--dark-gray-rgb), 0.5)` until we define with juani a standard for the opacity

src/intl/en.json
```diff
@@ -43,6 +43,13 @@
     "button": "Go Back"
   },
   "error": {
+    "message": {
+      "hide": "Hide error",
+      "show": "Show error",
+      "copy": "Copy to Clipboard",
+      "copied": "Copied!",
+      "call_for_action": "We rely on an external service that *can* fail. However, feel free to post this error message on [Discord](https://discord.com/channels/894658869391933540/987439629533855774)."
```
ncomerci, line 51: I would "We rely on external services that can fail..." (in plural)

ncomerci (approved):
🥇

src/components/Error/ErrorMessage.css
1emu: It's not defined in master :/

Maybe when you rebase you can replace the occurrences? 🙏

PR #387, opened 2022-06-23
Grants past items styles + new Pill component

andyesp:


1emu (approved):
😍😍😍

src/components/Icon/ChevronRight.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
 
-function ChevronRight() {
+function ChevronRight({ color }: { color?: string }) {
```
ncomerci, line 3: I think it isn't necessary to have `ChevronLeft` and `ChevronRight`. We can have a `Chevron` component that receives the left or right orientation by parameter and that becomes the svg property `transform="rotate(0)"` and `transform="rotate(180)"`.

For example: 
image

ncomerci (approved):

PR #410, opened 2022-07-14
fix: Update coauthoring status only when proposal is active

ncomerci:


andyesp (approved):

src/entities/Coauthor/routes.ts
```diff
@@ -16,16 +20,35 @@ export default routes((route) => {
   route.put('/coauthors/:proposal', withAuth, handleAPI(updateStatus))
 })
 
+async function filterCoauthorRequests(requests: CoauthorAttributes[]) {
+  const now = Date.now()
+  const filteredRequests: CoauthorAttributes[] = []
+  for (const request of requests) {
+    if (request.status === CoauthorStatus.PENDING) {
+      const proposal = await ProposalModel.findOne<ProposalAttributes>({ id: request.proposal_id, deleted: false })
+      if (proposal && now < proposal.finish_at.getTime()) {
```
1emu, line 29: We are repeating this check of the proposal date here and on the `updateStatus` function: 
`proposal && now < proposal.finish_at.getTime()`
`Date.now() < proposal.finish_at.getTime()`

(also in the front `const isUpdatable = Date.now() < proposalFinishDate.getTime()`

It's a simple logic but I wonder if we should encapsulate it in a function, at least to make the check more explicit in its intention.

Maybe in proposal utils:

`function coauthoringIsUpdatable(whatevs)`

src/entities/Proposal/model.ts
```diff
@@ -247,7 +252,10 @@ export default class ProposalModel extends Model<ProposalAttributes> {
         )}
     WHERE "deleted" = FALSE 
     ${conditional(!!filter.user && !filter.coauthor, SQL`AND p."user" = ${filter.user}`)} 
-    ${conditional(!!filter.coauthor, SQL`AND lower(c."address") = lower(${filter.user})`)}
+    ${conditional(
+      !!filter.coauthor,
+      SQL`AND lower(c."address") = lower(${filter.user}) AND (CASE WHEN p."finish_at" < NOW() THEN c."status" IN (${CoauthorStatus.APPROVED}, ${CoauthorStatus.REJECTED}) ELSE TRUE END)`
```
1emu, line 257: 🧠

1emu (approved):
LGTM! left a minor comment

src/entities/Proposal/model.ts
ncomerci: I don't think this part can be encapsulated since it depends on the finish date of each proposal 🤔

src/entities/Coauthor/utils.ts
```diff
@@ -0,0 +1,3 @@
+export function coauthoringIsUpdatable(proposalFinishDate: Date) {
```
andyesp, line 1: `isCoauthoringUpdatable`?

PR #413, opened 2022-07-15
feat: create pending updates when grant status is passed

andyesp:
⚠️ Change base to master after grants-dashboard merge.


1emu (approved):
👌 !

PR #418, opened 2022-07-18
feat: past grants list for mobile.

1emu:
image

image

image

image

### DIFF
--- a/src/components/Grants/PastGrantsTable.css
+++ b/src/components/Grants/PastGrantsTable.css
@@ -1,4 +1,4 @@
-.GrantsPage__PastGrantsTableHeader {
+.PastGrantsTable__Header {
   border: none !important;
   font-weight: 600 !important;
   font-size: 13px !important;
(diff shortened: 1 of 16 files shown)


src/components/Grants/PastGrantsTable.css
```diff
@@ -0,0 +1,25 @@
+.GrantsPage__PastGrantsTableHeader {
```
andyesp, line 1: In this file, we can rename this to `.PastGrantsTable__Header`, etc...

src/components/Icon/Sort.tsx
```diff
@@ -2,9 +2,10 @@ import React from 'react'
 
 type SortProps = {
   rotate?: number
+  color?: string
 }
 
-function Sort({ rotate = 0 }: SortProps) {
```
andyesp, line 7: I like this approach but maybe we can have a type matching the colors in our "design system". So it can receive "black-400" or Colors.Black400 as the color prop.

PR #441, opened 2022-07-25
feat: add metrics cards + open proposals

andyesp:
Pending:
- Voted state in open proposal card

image


src/components/Home/MetricsCards.tsx
```diff
@@ -0,0 +1,98 @@
...
+  pageDots: false,
+  prevNextButtons: false,
+  draggable: true,
+  dragThreshold: 10,
+  selectedAttraction: 0.01,
+  friction: 0.15,
+}
+
+const MetricsCards = () => {
+  const responsive = useResponsive()
```
ncomerci, line 29: we should stop using useResponsive as it's deprecated, there's a new hook (refactor fridays!)

src/components/Home/MetricsCards.tsx
```diff
@@ -0,0 +1,98 @@
...
+    () => [
+      {
+        category: t('page.home.metrics.proposals'),
+        title: t('page.home.metrics.active_proposals', { value: activeProposals?.total }),
+        description: t('page.home.metrics.ending_soon', { value: endingSoonProposals?.total }),
+      },
+      {
+        category: t('page.home.metrics.participation'),
+        title: t('page.home.metrics.votes_this_week', { value: 100 }),
+        description: t('page.home.metrics.votes_last_month', { value: 352 }),
```
ncomerci, line 64: You can use de the new `getVotes` method from Snapshot.ts

src/entities/Proposal/routes.ts
```diff
@@ -621,7 +623,7 @@ export async function updateProposalStatus(req: WithAuth<Request<{ proposal: str
   const user = req.auth!
   const id = req.params.proposal
   if (!isCommitee(user)) {
-    throw new RequestError('Only committed menbers can enact a proposal', RequestError.Forbidden)
+    throw new RequestError('Only committed members can enact a proposal', RequestError.Forbidden)
```
ncomerci, line 626: "committee members" ?

ncomerci (approved):
👏🏻👏🏻

src/hooks/useVotesCountByDate.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Snapshot } from './../api/Snapshot'
+
+export default function useVotesCountByDate(start: Date, end: Date) {
+  const [votes, state] = useAsyncMemo(
+    async () => {
+      return await Snapshot.get().getVotes(start, end)
+    },
+    [start, end],
+    {
+      callWithTruthyDeps: true,
```
ncomerci, line 12: Maybe initialValue = 0 ?

src/api/Snapshot.ts
```diff
@@ -455,15 +455,15 @@ export class Snapshot extends API {
   }
 
   fetchProposals = async (
-    params: { start: Date; end: Date; fields: (keyof SnapshotProposal)[] },
+    params: { start: Date; end: Date; orderBy?: string; scoresState?: string; fields: (keyof SnapshotProposal)[] },
```
ncomerci, line 458: I think it would be good to constrain the values that scoresState can take instead of 'string' 🤔

src/pages/index.tsx
```diff
@@ -40,6 +42,10 @@ export default function HomePage() {
       <Navigation activeTab={NavigationTab.Home} />
       <Container>
         <MainBanner />
+      </Container>
+      <MetricsCards />
```
ncomerci, line 46: Is this container really necessary? I think it would be cleaner if we just use a single container that contains all the components of the home page

ncomerci (commented):
Sorry for the delay! 🥲

src/pages/index.tsx
andyesp: We need to render the Container conditionally in this component as it is not used in mobile.
image

PR #447, opened 2022-07-27
Community engagement

ncomerci:
### Desktop
Captura de Pantalla 2022-08-01 a la(s) 12 42 45
Captura de Pantalla 2022-08-01 a la(s) 12 42 51

### Mobile
localhost_8000_ (1)


src/components/HomeCharts/HomeCharts.tsx
```diff
@@ -0,0 +1,76 @@
...
+            {t('page.home.community_engagement.votes_per_proposal_title')}
+          </Tabs.Tab>
+        </Tabs.Left>
+        <Tabs.Right>
+          <Header sub className="Display">
+            {t('page.home.community_engagement.display_median')}
+          </Header>
+        </Tabs.Right>
+      </Tabs>
+      {Charts[selectedTab]}
```
andyesp, line 71: I'd use conditional rendering here to avoid generating an object on the render.

```
{selectedTab === ChartType.ParticipatingVP &&  <LineChart ... />}
{selectedTab === ChartType.VotesPerProposal &&  <LineChart ...}
```

src/hooks/useParticipatingVP.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { Snapshot, SnapshotProposal } from '../api/Snapshot'
+import { SNAPSHOT_SPACE } from '../entities/Snapshot/constants'
+import { groupProposalsByMonth, median } from '../entities/Snapshot/utils'
+
+export default function useParticipatingVP(start: Date, end: Date) {
+  const [proposals, state] = useAsyncMemo(
+    async () => {
+      // TODO: un-hardcode snapshot space
+      return await Snapshot.get().getProposals('snapshot.dcl.eth', start, end, ['created', 'scores_total'])
```
andyesp, line 13: It'd be great to update getProposals to not receive the space as an argument and use it from the constants file.

src/hooks/useTopVoters.ts
```diff
@@ -0,0 +1,42 @@
...
+type Voters = {
+  address: string
+  votes: number
+}
+
+export default function useTopVoters(start: Date, end: Date, limit: number) {
+  const [votes, state] = useAsyncMemo(
+    async () => {
+      // TODO: un-hardcode snapshot space
+      const votes = await Snapshot.get().getVotes('snapshot.dcl.eth', start, end)
```
andyesp, line 18: Here too

src/hooks/useParticipatingVP.ts
ncomerci: TODO in a new PR as discussed

andyesp (changes requested):

src/hooks/useAbbreviatedNumber.ts
```diff
@@ -1,22 +1,29 @@
 import useFormatMessage from 'decentraland-gatsby/dist/hooks/useFormatMessage'
 
-export default function useAbbreviatedNumber(value: number) {
+export default function useAbbreviatedNumber() {
```
andyesp, line 3: I prefer if this receives the value and returns the value formatted.

But if you really need this to return the formatter I'd rename it to useAbbreviatedFormatter or similar... 🤔

src/entities/Snapshot/utils.ts
```diff
@@ -45,3 +45,39 @@ export function calculateMatch(votes1: SnapshotVote[] | null, votes2: SnapshotVo
...
+    return array[half]
+  }
+
+  return (array[half - 1] + array[half]) / 2.0
+}
+
+export function groupProposalsByMonth(proposals: Partial<SnapshotProposal>[], field: keyof SnapshotProposal) {
+  const data: Record<string, number[]> = {}
+
+  for (const proposal of proposals) {
```
1emu, line 65: I like the creation date validation.
I'm thinking it might be event better to catch the failing proposal inside the `for` loop and don't throw an error but rather log and continue processing the others (if that makes sense):

```typescript
for (const proposal of proposals) {
  if (proposal.created) {
       const proposalDate = new Date(proposal.created * 1000)
       const month = proposalDate.getMonth()
       const year = proposalDate.getFullYear()
       const key = `${year}/${month + 1}`
       const value = proposal[field]
       if (typeof value === 'number') {
          data[key] = [...(data[key] || []), value]
       } else {
         console.log(`Proposal field is not a number: ${value}`)  // TODO: rollbar/newrelic
       }
    } else {
      console.log('Proposal has no creation date') // TODO: rollbar/newrelic
    }
 }
```

src/api/Snapshot.ts
```diff
@@ -341,6 +378,102 @@ export class Snapshot extends API {
     const vp = await this.getLatestScores(space, [address])
     return Object.values(vp)[0]
   }
+
+  async getVotes(space: string, start: Date, end: Date) {
```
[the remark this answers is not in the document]
1emu: Also, I think we shouldn't need the space as an argument, right? The Snapshot API class already knows the space based on the env, and we aren't querying more than one space per environment.

It could have it inside the method:

```typescript
 variables: {
              space: SNAPSHOT_SPACE,
              start: getQueryTimestamp(start.getTime()),
              end: getQueryTimestamp(end.getTime()),
              first,
              skip,
            },
            ```

File: src/hooks/useTopVoters.ts [hunk 7]
```diff
@@ -0,0 +1,41 @@
...
+      const votes = await Snapshot.get().getVotes(SNAPSHOT_SPACE, start, end)
+      const votesByUser = votes.reduce((acc, vote) => {
+        const address = vote.voter.toLowerCase()
+        acc[address] = (acc[address] || 0) + 1
+        return acc
+      }, {} as Record<string, number>)
+
+      return Object.entries(votesByUser).map<Voters>(([address, votes]) => ({ address, votes }))
+    },
+    [limit],
```
1emu, line 26: I don't think the asyncMemo should re-evaluate depending on the limit change, because you are not using it to fetch different data (it only depends on the start and end date). And limit is only used in the slice at the end of the useTopVoters hook.

1emu (approved):
Lovely! Left a few suggestions :)

andyesp (approved):

PR #448, opened 2022-07-27
fix: show updates on enacted grants and use last 7 chars of snapshot id on discourse title

andyesp:
image


1emu (approved):
This makes me think we should add some tests for this new "displayable updates" condition...

PR #454, opened 2022-08-02
feat: remove threshold to post updates

andyesp:


1emu (approved):
This got me thinking if we should't have any kind of validation for updates creation in the backend... 🤔

PR #465, opened 2022-08-12
fix: line charts date display bug in firefox

1emu:
Firefox is not supporting date parsing without the day

image

image

image


andyesp (approved):
good catch!

PR #498, opened 2022-09-02
chore: update next workflows

andyesp:


.github/workflows/next_pull_request_build.yml
```diff
@@ -9,41 +9,10 @@ on:
...
-
-      - name: building
-        run: npm run build
+  build:
+    uses: decentraland/actions/.github/workflows/build-quay-main.yml@main
+    with:
+      service-name: governance
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
```
menduz, line 18: this whole file shouldn't be necessary anymore I think

.github/workflows/next_build.yml
```diff
@@ -1,46 +1,15 @@
...
-
-      - name: building
-        run: npm run build
+  build:
+    uses: decentraland/actions/.github/workflows/build-quay-main.yml@main
+    with:
+      service-name: governance
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
```
menduz, line 15: I think you are forgetting to deploy the image to a specific env https://github.com/decentraland/sdk-content-server/blob/main/.github/workflows/docker.yml#L13

andyesp: Done! So if we want to deploy to all 3 environments we need to create a new workflow for each one. Right?

.github/workflows/next_pull_request_build.yml
andyesp: Removed. It'd be nice to have deployments per pull request but we can review that later after the deploy on merge is working ok as it probably needs a different setup.

.github/workflows/next_build.yml
menduz: separate them by space `dev stg prd`

ncomerci (approved):

PR #505, opened 2022-09-07
feat: profile box component

1emu:
First version for this component, we can add things to it as we need.

image

### DIFF
--- a/src/components/Profile/ProfileBox.css
+++ b/src/components/Profile/ProfileBox.css
@@ -1,0 +1,33 @@
+.ProfileBox__Container {
+  background: var(--white-900);
+  border: 1px solid var(--black-100);
+  box-shadow: 0 1px 8px var(--black-100);
+  border-radius: 8px;
+  padding-top: 14px;
+  padding-bottom: 16px;
+}
+
+.ProfileBox__Padded {
+  padding-right: 16px;
+  padding-left: 16px;
+}
+
+.ProfileBox__Header {
+  display: flex;
+}
+
+.ProfileBox__Header > span {
+  font-weight: 500;
+  font-size: 13px;
+  line-height: 18px;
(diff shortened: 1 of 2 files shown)


src/components/Profile/ProfileBox.css
```diff
@@ -0,0 +1,33 @@
...
+  font-weight: 500;
+  font-size: 13px;
+  line-height: 18px;
+  text-transform: uppercase;
+  color: var(--black-600);
+  margin-bottom: 13px;
+  margin-right: 4px;
+}
+
+.ProfileBox__Divider hr.Divider {
```
andyesp, line 29: Is it necessary to add this `hr.Divider`?

ncomerci (approved):

src/components/Profile/ProfileBox.css
1emu: We can write it as `.Divider`, but it's necessary because the custom class name is applied to the `<hr>` parent

```tsx
<div className={TokenList.join([size && `Divider__Container--${size}`, className])}>
      <hr className="Divider" />
    </div>
    ```

andyesp: And can we move the margins to a Divider__Container class? And change the component to receive a color?

andyesp: Also border style is what updates the divider color right? What if we pass the color as a prop instead of overriding it?

1emu: > Also border style is what updates the divider color right? What if we pass the color as a prop instead of overriding it?

I'm adding this in the next PR. 

> And can we move the margins to a Divider__Container class so the className override affects this margin? And change the component to receive a color?



> Also border style is what updates the divider color right? What if we pass the color as a prop instead of overriding it?

I'm adding this in the next PR

PR #517, opened 2022-09-15
Voting Power Distribution

1emu:
Replaced get scores with new voting power distribution snapshot endpoint


src/clients/SnapshotGraphql.ts
```diff
@@ -374,4 +394,48 @@ export class SnapshotGraphql extends API {
...
+    )
+
+    if (!result?.data?.vp) {
+      throw Error('Unable to fetch VP Distribution')
+    }
+
+    const vpByStrategy = result?.data?.vp.vp_by_strategy
+
+    return {
+      totalVp: Math.floor(result?.data?.vp.vp),
```
andyesp, line 430: What do you think about dropping `Vp` from all properties names? `{ own, wMana, land, estate, mana, names, delegated, linkedWearables }`

src/clients/SnapshotGraphql.ts
```diff
@@ -374,4 +394,48 @@ export class SnapshotGraphql extends API {
...
+      ownVp: Math.floor(result?.data?.vp.vp - vpByStrategy[5]),
+      wManaVp: Math.floor(vpByStrategy[0]),
```
andyesp, line 432: We can have an index mapper... something like... IDK if we are going to use or we are using this strategy order somewhere else already.

```
enum StrategyOrder {
 WrappedMana,
 Land,
 Estate,
 ...
}

... 
vpByStrategy[StrategyOrder.WrappedMana]
```

src/components/Modal/VotingPowerDelegationDetail/VotingPowerDistribution.tsx
```diff
@@ -3,26 +3,37 @@ import React from 'react'
...
+
   const t = useFormatMessage()
-  const total = mana + name + land + delegated
+  const total = vpDistribution.totalVp
+  const mana = vpDistribution.manaVp
+  const name = vpDistribution.namesVp
+  const land = vpDistribution.landVp
+  const delegated = vpDistribution.delegatedVp
+  const estate = vpDistribution.estateVp
+  const linkedWearables = vpDistribution.linkedWearablesVp
```
andyesp, line 29: `const { total, mana, name, land, delegated, estate, linkedWearables } = vpDistribution` :)

src/components/Modal/VotingPowerDelegationDetail/VotingPowerDistribution.tsx
```diff
@@ -3,26 +3,37 @@ import React from 'react'
...
-  delegated: number
+  vpDistribution?: VpDistribution | null
   className?: string
 }
 
 const getPercentage = (value: number, total: number): string => `${((value * 100) / total).toFixed(2)}%`
 
-const VotingPowerDistribution = ({ mana, name, land, delegated, className }: Props) => {
+const VotingPowerDistribution = ({ vpDistribution, className }: Props) => {
+  if (!vpDistribution) vpDistribution = EMPTY_DISTRIBUTION
```
andyesp, line 20: I'd avoid reassigning props. We can add `vpDistribution={vpDistribution || EMPTY_DESTRIBUTION}` as a prop where this component is rendered.

andyesp: If we are getting this value from the hook, is it necessary to add a fallback here as the hook has a fallback?

src/clients/SnapshotGraphql.ts
1emu: Love this. Let's make it explicit

andyesp (approved):

PR #528, opened 2022-09-20
fix: token symbol in grants

andyesp:


src/clients/DclData.ts
```diff
@@ -45,7 +45,10 @@ type Grants = {
   status: string
   symbol: string
```
1emu, line 46: is this symbol still present?

1emu (approved):

ncomerci (approved):

src/clients/DclData.ts
andyesp: no! I'll remove it

PR #529, opened 2022-09-20
New profile grants card

ncomerci:
image


src/components/Grants/GrantCard/ProfileGrantCard.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import ProfileGrantItem from './ProfileGrantItem'
+
+interface Props {
+  grants: GrantAttributes[]
+}
+
+const MAX_GRANTS = 4
+
+function ProfileGrantCard({ grants }: Props) {
```
1emu, line 14: This is subtle, so I open the discussion. 
I think i'd rename this to `ProfileGrantCards`, or `ProfileGrants` or `ProfileGrantsSection`, and each `ProfileGrantItem` to `ProfileGrantCard`.

Just because on the other pages (home and grants) we have cards for grants, and there's a one to one relationship between a card and a grant. 

It seems strange for me to have a card with several grants inside. 

WDYT?

src/components/Grants/GrantCard/ProgressBarTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  if (isInCliff) {
+    const now = Time.utc()
+    const vestingStartDate = Time.unix(enacted_at)
+    const elapsedSinceVestingStarted = now.diff(vestingStartDate, 'day')
+    const daysToGo = CLIFF_PERIOD_IN_DAYS - elapsedSinceVestingStarted
+
+    textToShow = t('page.profile.grants.cliff_period', { count: daysToGo })
+  } else if (isOTP) {
+    textToShow = t('page.profile.grants.one_time_tx', { time: formatDate(new Date(enacted_at * 1000)) })
```
1emu, line 36: same here for `enacted_at`

src/components/Grants/GrantCard/VestingProgress.tsx
```diff
@@ -64,18 +69,20 @@ const VestingProgress = ({ grant }: Props) => {
...
-          <span className="VestingProgress__VestedDate">{enactedDate}</span>
-        </div>
-        {contract?.finish_at && (
+      {!basic && (
+        <div className="VestingProgress__Dates">
           <div className="VestingProgress__VestedAt">
-            <span>{t('page.grants.end_date')}</span>
-            <span className="VestingProgress__VestedDate">{Time.unix(contract.finish_at).fromNow()}</span>
+            <span>{enacting_tx ? t('page.grants.transaction_date') : t('page.grants.started_date')}</span>
+            <span className="VestingProgress__VestedDate">{enactedDate}</span>
```
1emu, line 76: here we are also using `const enactedDate = Time.unix(enacted_at).fromNow()`

src/entities/Proposal/utils.ts
```diff
@@ -212,3 +214,11 @@ export const isCoauthor = async (proposalId: string, address: string): Promise<b
     .getCoAuthorsByProposal(proposalId, CoauthorStatus.APPROVED)
     .then((coauthors) => !!coauthors.find((coauthor) => coauthor.address === address))
 }
+
+const TRANSPARENCY_ONE_TIME_PAYMENT_TIERS = new Set(['Tier 1', 'Tier 2'])
+
+export function isProposalInCliffPeriod(grant: GrantAttributes) {
+  const isOneTimePayment = TRANSPARENCY_ONE_TIME_PAYMENT_TIERS.has(grant.configuration.tier)
+  const now = Time.utc()
+  return !isOneTimePayment && Time.unix(grant.enacted_at).add(CLIFF_PERIOD_IN_DAYS, 'day').isAfter(now)
```
1emu, line 223: `const isOneTimePayment = TRANSPARENCY_ONE_TIME_PAYMENT_TIERS.has(grant.configuration.tier)`

This could be another question the object could know how to reply: `grant.isOneTimePayment()`

and `grant.isInCliffPeriod()`

1emu (approved):
Nice! 
Left some comments for us to discuss :)

src/components/Grants/GrantCard/ProfileGrantCard.tsx
andyesp: `ProfileGrantList`?

src/components/Grants/GrantCard/ProfileGrantItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+interface Props {
+  grant: GrantAttributes
+}
+
+function ProfileGrantItem({ grant }: Props) {
+  const t = useFormatMessage()
+  const intl = useIntl()
+  const { user, title, enacted_at, token } = grant
+  const enactedDate = new Date(enacted_at * 1000)
```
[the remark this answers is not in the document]
andyesp: I'd move `value => Time.unix(value).fromNow()` to a helper for now. (`helpers/grants.ts`?)

Not sure how the Grant class is going to work. Are we going to return a Grant from the Governance client?

src/components/Grants/GrantCard/ProfileGrantItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+      <div className="ProfileGrantItem__VestingSection">
+        <div className="ProfileGrantItem__PillContainer">
+          <GrantPill type={grant.configuration.category} />
+        </div>
+        <div className="ProfileGrantItem__VestingProgressContainer">
+          <ProgressBarTooltip grant={grant} isInCliff={proposalInCliffPeriod}>
+            <div>
+              {proposalInCliffPeriod ? (
+                <CliffProgress enactedAt={grant.enacted_at} basic />
```
andyesp, line 57: You can use `enacted_at` here as grant is already destructured.

src/components/Grants/GrantCard/ProgressBarTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+  isInCliff: boolean
+  children: React.ReactNode
+}
+
+function ProgressBarTooltip({ grant, isInCliff, children }: Props) {
+  const t = useFormatMessage()
+  const intl = useIntl()
+
+  const { contract, tx_amount, token, enacted_at } = grant
+  const isOTP = !contract
```
andyesp, line 22: What is OTP?

andyesp: Oh one time Payment... Maybe `isOneTimePayment`

src/components/Grants/GrantCard/ProgressBarTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const vestedAmount = (contract ? contract.vestedAmount : tx_amount) || 0
+  const releasedAmount = !isOTP ? contract.released : 0
+
+  let textToShow = ''
```
andyesp, line 26: This can be in a helper function to make this component easier to read
`const text = getText(...)`
But maybe it will have too many arguments...

src/pages/profile.tsx
```diff
@@ -25,7 +29,10 @@ export default function ProfilePage() {
   const address = isEthereumAddress(params.get('address') || '') ? params.get('address') : userAddress
   const isLoggedUserProfile = userAddress === address
 
-  const { delegation, delegationState, scores, isLoadingScores, ownVotingPower } = useVotingPowerInformation(address)
+  const [grants] = useGrantsByUser(address!, true)
+  const { delegation, delegationState, scores, isLoadingScores, vpDistribution } = useVotingPowerInformation(address)
+
+  const areGrants = useMemo(() => grants.total > 0, [grants.total])
```
andyesp, line 35: `hasGrants`

src/pages/profile.tsx
```diff
@@ -50,24 +57,30 @@ export default function ProfilePage() {
...
       <Head
         title={t('page.profile.title') || ''}
         description={t('page.profile.description') || ''}
         image="https://decentraland.org/images/decentraland.png"
       />
       <Navigation activeTab={NavigationTab.Profile} />
       <Container className="Profile__Container">
+        {areGrants && (
+          <ProfileBox title={t('page.profile.grants.title')}>
+            <ProfileGrantCard grants={[...grants.current, ...grants.past]} />
```
andyesp, line 70: The `useGrantsByUser` can return this array.

src/pages/profile.tsx
```diff
@@ -25,7 +29,10 @@ export default function ProfilePage() {
   const address = isEthereumAddress(params.get('address') || '') ? params.get('address') : userAddress
   const isLoggedUserProfile = userAddress === address
 
-  const { delegation, delegationState, scores, isLoadingScores, ownVotingPower } = useVotingPowerInformation(address)
+  const [grants] = useGrantsByUser(address!, true)
```
andyesp, line 32: What if there is no user? I wouldn't force the `!` and make the hook handle an undefined case.

src/hooks/useGrantsByUser.ts
```diff
@@ -0,0 +1,12 @@
+import useAsyncMemo from 'decentraland-gatsby/dist/hooks/useAsyncMemo'
+
+import { Governance } from '../clients/Governance'
+import { GrantsResponse } from '../entities/Proposal/types'
+
+export default function useGrantsByUser(address: string, coauthoring?: boolean) {
+  return useAsyncMemo(
+    async () => await Governance.get().getGrantsByUser(address, coauthoring),
```
andyesp, line 8: In what case `coauthoring` is false?

src/pages/profile.tsx
andyesp: Also I don't think is necessary to memoize this condition.

src/components/Grants/GrantCard/ProfileGrantItem.tsx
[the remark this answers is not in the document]
1emu: I was thinking on wrapping the Grants that the governance client returns somewhere in the frontend, and having a frontend only object (because methods associated with it will be related to how the object displays/shows it's info)

The advantage to having this kinds of functions scattered on util/helper files is that you don't need to remember they exist.
When you add the methods to the object, then the only way to fetch the data you need to use is through the object's interface.

src/components/Common/SkeletonBars.tsx
```diff
@@ -0,0 +1,20 @@
...
+import './SkeletonBars.css'
+
+interface Props {
+  height: number
+  amount: number
+}
+
+function SkeletonBars({ amount, height }: Props) {
+  // eslint-disable-next-line prefer-spread
+  const bars = Array.apply(null, Array(amount)).map((_, i) => (
```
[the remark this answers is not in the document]
ncomerci: it doesn't work ☹️

src/components/Grants/GrantCard/ProgressBarTooltip.tsx
ncomerci: yep, I don't think it's worth it. It has too many arguments...

src/hooks/useGrantsByUser.ts
ncomerci: For the time being, never. I left it ready in case we need that option in the future.

src/components/Common/SkeletonBars.tsx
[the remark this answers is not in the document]
andyesp: I think lodash has a `times` function that can work here. Otherwise I think the JS way is better but don't like the eslint-disable though :/

src/pages/profile.tsx
andyesp: But we can see other users profile right? We should be able to view those logged out.

andyesp (approved):

src/components/Common/SkeletonBars.tsx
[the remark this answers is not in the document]
ncomerci: `times` worked perfectly, thanks!

src/pages/profile.tsx
ncomerci: You are right, fixed it

PR #530, opened 2022-09-21
fix: snapshot delegations query limit 

andyesp:
Tested with user 0xd5e9ef1cedad0d135d543d286a2c190b16cbb89e.

Before:
image

After:
image

### DIFF
--- a/src/entities/Snapshot/queries.ts
+++ b/src/entities/Snapshot/queries.ts
@@ -1,32 +1,22 @@
-export const DELEGATIONS_ON_PROPOSAL_QUERY = `
-query ($space: String!, $address: String!, $blockNumber: Int) {
-  delegatedTo: delegations(block:{number: $blockNumber},where: { space_in: ["", $space], delegator: $address }, orderBy: timestamp, orderDirection: desc) {
-    delegator
-    delegate
-    space
-    timestamp
-  },
-  delegatedFrom: delegations(block:{number: $blockNumber},where: { space_in: ["", $space], delegate: $address }, orderBy: timestamp, orderDirection: desc) {
+type BlockNumber = string | number | undefined
+const getBlockNumberArgument = (blockNumber: BlockNumber) => (blockNumber ? '$blockNumber: Int' : '')
+const getBlockNumberFilter = (blockNumber: BlockNumber) => (blockNumber ? 'block:{number: $blockNumber},' : '')
+const getDelegationType = (key: 'delegatedTo' | 'delegatedFrom') => {
+  if (key === 'delegatedTo') {
+    return 'delegator'
+  }
+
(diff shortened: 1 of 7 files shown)


src/entities/Snapshot/queries.ts
```diff
@@ -1,32 +1,22 @@
...
-    space
-    timestamp
-  },
-  delegatedFrom: delegations(block:{number: $blockNumber},where: { space_in: ["", $space], delegate: $address }, orderBy: timestamp, orderDirection: desc) {
-    delegator
-    delegate
-    space
-    timestamp
+type BlockNumber = string | number | undefined
+const getBlockNumberArgument = (blockNumber: BlockNumber) => (blockNumber ? '$blockNumber: Int' : '')
```
1emu, line 2: I think it's not necessary to return this as an empty string, because the `Int` without a bang `!` means it can be undefined.

Might just always work with `'$blockNumber: Int'`

src/entities/Snapshot/queries.ts
```diff
@@ -1,32 +1,22 @@
...
-    delegator
-    delegate
-    space
-    timestamp
-  },
-  delegatedFrom: delegations(where: { space_in: ["", $space], delegate: $address }, orderBy: timestamp, orderDirection: desc) {
+  return 'delegate'
+}
+
+export const getDelegatedQuery = (key: 'delegatedTo' | 'delegatedFrom', blockNumber?: BlockNumber) => `
```
1emu, line 12: Niiice 😎

1emu (approved):
🪄

ncomerci (approved):

src/entities/Snapshot/utils.ts
```diff
@@ -124,66 +124,63 @@ export function filterDelegationFrom(delegations: Delegation[], space: string):
...
+  return Array.from(uniqueDelegations.values())
 }
 
-function getFetchDelegatesVariables(address: string, blockNumber: string | number) {
-  const variables: any = {
+function getDelegatesVariables(address: string, blockNumber?: string | number) {
+  return {
     address: address.toLowerCase(),
     space: SNAPSHOT_SPACE,
+    ...(!blockNumber && { blockNumber }),
```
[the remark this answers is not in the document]
andyesp: You are rightttt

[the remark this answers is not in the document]
andyesp: I think the first one is ok

src/entities/Snapshot/queries.ts
andyesp: True! 🙌

1emu (approved):

ncomerci (approved):

PR #534, opened 2022-09-23
fix: fetch coauthor grants from governance api

andyesp:


src/entities/Proposal/routes.ts
```diff
@@ -620,7 +621,8 @@ async function getGrantsByUser(req: Request): ReturnType<typeof getGrants> {
   let coauthoringProposalIds = new Set<string>()
 
   if (isCoauthoring) {
-    const coauthoring = await Governance.get().getProposalsByCoAuthor(address, CoauthorStatus.APPROVED)
+    const requests = await CoauthorModel.findProposals(address, CoauthorStatus.APPROVED)
+    const coauthoring = await filterCoauthorRequests(requests)
```
ncomerci, line 625: I think it's not necessary to filter the requests due to they are only the approved ones 

image

ncomerci (approved):

PR #548, opened 2022-10-03
User voting stats

1emu:
Hide Personal Match when user is not logged in
image

Show Personal Match if you are logged in and are looking at your profile
image

Show Personal Match if logged in and looking at another user's profile
image
image

Profile with no votes or VP
image

Comparison with user that has old votes (and you did not voted on the same proposals)
image


src/components/User/UserVotingStats.tsx
```diff
@@ -0,0 +1,93 @@
...
+    ? 'page.profile.user_voting_stats.outcome_match_conclusion_<50'
+    : outcomeMatchPercentage <= 80
+    ? 'page.profile.user_voting_stats.outcome_match_conclusion_50-80'
+    : 'page.profile.user_voting_stats.outcome_match_conclusion_>80'
+}
+
+export default function UserVotingStats({ address }: Props) {
+  const t = useFormatMessage()
+  const [userAddress] = useAuthContext()
+  const sameUser =
```
andyesp, line 39: Are we ready to move this to a `isSameAddress` helper?

src/entities/Snapshot/utils.ts
```diff
@@ -54,6 +54,32 @@ export function calculateMatch(votes1: SnapshotVote[] | null, votes2: SnapshotVo
...
+  let closedProposalsCounter = 0
+  for (const vote of votes) {
+    if (vote.proposal?.state === 'closed' && vote.proposal?.scores && vote.proposal?.scores.length > 0) {
+      closedProposalsCounter++
+      const scores = vote.proposal.scores
+      const winnerChoiceScore = Math.max(...scores)
+      const winnerChoiceIndex = scores.indexOf(winnerChoiceScore)
+      const resultIsNotATie = scores.filter((score) => score === winnerChoiceScore).length === 1
+      if (resultIsNotATie) {
+        if (vote.choice === winnerChoiceIndex + 1) {
```
andyesp, line 73: I think this can be in one condition

`if (resultIsNotATie && vote.choice === winnerChoiceIndex + 1) {`

src/hooks/useVotingStats.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (userAddress && getChecksumAddress(vote.voter!) === getChecksumAddress(userAddress!)) {
+      userVotes.push(vote)
+    } else {
+      addressVotes.push(vote)
+    }
+  })
+  return { addressVotes, userVotes }
+}
+
+function getOutcomeMatchPercentage(addressVotes: SnapshotVote[]) {
```
andyesp, line 21: I'd remove this function and do the addresses check on the hook return directly.

src/hooks/useVotingStats.ts
```diff
@@ -0,0 +1,70 @@
...
+    async () => {
+      return await SnapshotGraphql.get().getProposals(aMonthAgo, now, ['id', 'choices', 'scores'])
+    },
+    [],
+    { initialValue: [] as Partial<SnapshotProposal>[], callWithTruthyDeps: true }
+  )
+
+  const [votes, votesState] = useAsyncMemo(
+    async () => {
+      const addresses = [address]
```
andyesp, line 51: In one line and without an if like...
`const addresses = [address, userAddress].filter(element => !!element)`
or
`const addresses = [address, userAddress].filter(Boolean)`

🤷‍♂️

src/hooks/useVotingStats.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (userAddress) addresses.push(userAddress)
+      return await SnapshotGraphql.get().getAddressesVotes(addresses)
+    },
+    [address],
+    { initialValue: [] as SnapshotVote[], callWithTruthyDeps: true }
+  )
+  const { addressVotes, userVotes } = sortAddressesVotes(votes, userAddress)
```
andyesp, line 58: Maybe we should useMemo this one.

src/hooks/useVotingStats.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const { participationTotal, participationPercentage } = getParticipation(last30DaysProposals, addressVotes, aMonthAgo)
+  const matchResult = calculateMatch(addressVotes, userVotes)
```
andyesp, line 61: And memo these ones too... 🤔

src/hooks/useVotingStats.ts
1emu: I can't, it complains that addresses is a `(string|null)[]` :(

andyesp (approved):

PR #560, opened 2022-10-06
Delegated VP card & minor fixes

ncomerci:
image

### DIFF
--- a/src/components/Profile/VpDelegationBox.tsx
+++ b/src/components/Profile/VpDelegationBox.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
 
-import { AsyncStateResultState } from 'decentraland-gatsby/dist/hooks/useAsyncState'
 import useFormatMessage from 'decentraland-gatsby/dist/hooks/useFormatMessage'
 import { Container } from 'decentraland-ui/dist/components/Container/Container'
 import Grid from 'semantic-ui-react/dist/commonjs/collections/Grid/Grid'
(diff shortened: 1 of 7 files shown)


src/components/Profile/VpDelegationBox.tsx
```diff
@@ -0,0 +1,59 @@
...
+import SkeletonBars from '../Common/SkeletonBars'
+import DelegatorCardProfile from '../Delegation/DelegatorCardProfile'
+import Scale from '../Icon/Scale'
+
+import { ProfileBox } from './ProfileBox'
+import './VpDelegationBox.css'
+
+interface Props {
+  delegation: DelegationResult
+  delegationState: AsyncStateResultState<DelegationResult, DelegationResult>
```
andyesp, line 19: If you are only using loading I'd simplify this to `isLoadingDelegations: boolean`

andyesp (approved):

src/components/Profile/VpDelegatorsBox.tsx
```diff
@@ -30,7 +30,7 @@ export default function VpDelegatorsBox({
 }: Props) {
   const t = useFormatMessage()
   const [userAddress] = useAuthContext()
-  const isLoggedUserProfile = userAddress === address
+  const isLoggedUserProfile = userAddress?.toLowerCase() === address?.toLowerCase()
```
1emu, line 33: this is tricky... if both can be null/undefined, it could return true.

You can use `isSameAddress(userAddress, address)` :)

1emu (approved):

1emu:
> image

I think the bottom margin below the card should be the same as the top one (16px?)

src/components/Profile/VpDelegatorsBox.tsx
ncomerci: Thank you! I forgot about that

PR #582, opened 2022-10-24
feat: Voted proposals on profile

ncomerci:
## Missing
- [x] Helper text

### Screenshots
image
image

### DIFF
--- a/src/clients/SnapshotGraphqlTypes.ts
+++ b/src/clients/SnapshotGraphqlTypes.ts
@@ -21,7 +21,7 @@ export type SnapshotSpace = {
 }
 export type SnapshotVoteResponse = SnapshotQueryResponse<{ votes: SnapshotVote[] }>
 export type SnapshotVote = {
-  id?: string
+  id: string
   voter: string
   created: number
   vp?: number
--- a/src/clients/SnapshotGraphqlTypes.ts
+++ b/src/clients/SnapshotGraphqlTypes.ts
@@ -21,7 +21,7 @@ export type SnapshotSpace = {
 }
 export type SnapshotVoteResponse = SnapshotQueryResponse<{ votes: SnapshotVote[] }>
 export type SnapshotVote = {
-  id: string
+  id?: string
   voter: string
   created: number
   vp?: number
--- a/src/components/Profile/VotedProposalsBox.tsx
+++ b/src/components/Profile/VotedProposalsBox.tsx
@@ -1,16 +1,27 @@
-import React from 'react'
+import React, { useState } from 'react'
 
 import useFormatMessage from 'decentraland-gatsby/dist/hooks/useFormatMessage'
 import { Container } from 'decentraland-ui/dist/components/Container/Container'
 
+import useVotedProposals from '../../hooks/useVotedProposals'
+import ProfileProposalItem from '../Proposal/ProfileProposalItem'
+
 import { ProfileBox } from './ProfileBox'
 
-function VotedProposalsBox() {
+interface Props {
+  address: string
+}
+
+function VotedProposalsBox({ address }: Props) {
   const t = useFormatMessage()
+  const [fetchInfo, setFetchInfo] = useState({ first: 5, skip: 0 })
+  const { votes } = useVotedProposals(address, fetchInfo.first, fetchInfo.skip)
   return (
     <Container>
       <ProfileBox title={t('page.profile.voted_proposals.title')} info={t('page.profile.voted_proposals.helper')}>
-        <div>Proposals</div>
+        {votes.map((vote, index) => {
+          return <ProfileProposalItem key={vote.id} votedProposal={vote} />
+        })}
       </ProfileBox>
     </Container>
   )
--- a/src/components/Profile/VotedProposalsBox.tsx
+++ b/src/components/Profile/VotedProposalsBox.tsx
@@ -1,10 +1,11 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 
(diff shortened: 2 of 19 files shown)


src/components/Profile/VotedProposalsBox.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface Props {
+  address: string
+}
+
+const PROPOSALS_AMOUNT = 5
+
+function VotedProposalsBox({ address }: Props) {
+  const t = useFormatMessage()
+  const [skip, setSkip] = useState(0)
+  const { votes } = useVotedProposals(address, PROPOSALS_AMOUNT, skip)
```
andyesp, line 22: What if this hook returns a `viewMore` function and `votes` can handle this update so you can avoid doing this logic here? :)

src/intl/en.json
```diff
@@ -800,6 +800,15 @@
         "delegated": "Delegated {vp} VP",
         "delegate_action": "Delegate them your VP",
         "button": "Show all delegators"
+      },
+      "voted_proposals": {
+        "title": "Proposals voted on",
+        "helper": "...",
```
andyesp, line 806: This helper is missing right? Maybe we should add a `// TODO:`

src/clients/SnapshotGraphqlTypes.ts
```diff
@@ -21,7 +21,7 @@ export type SnapshotSpace = {
 }
 export type SnapshotVoteResponse = SnapshotQueryResponse<{ votes: SnapshotVote[] }>
 export type SnapshotVote = {
-  id?: string
+  id: string
```
1emu, line 24: This should remain optional unless we make sure that every query sent to graphql returns the vote `id`

src/components/Profile/VotedProposalsBox.tsx
```diff
@@ -0,0 +1,46 @@
...
+import ProfileProposalItem from '../Proposal/ProfileProposalItem'
+
+import { ProfileBox } from './ProfileBox'
+import './VotedProposalsBox.css'
+
+interface Props {
+  address: string
+}
+
+const PROPOSALS_AMOUNT = 5
```
1emu, line 17: I think we can call this proposals per page? or proposals displayed?

src/components/Profile/VotedProposalsBox.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+function VotedProposalsBox({ address }: Props) {
+  const t = useFormatMessage()
+  const [skip, setSkip] = useState(0)
+  const { votes } = useVotedProposals(address, PROPOSALS_AMOUNT, skip)
+  const [displayVotes, setDisplayVotes] = useState(votes)
```
1emu, line 23: I think it's better to call this `votesDisplayed` (I'd call it `displayVotes` in case it was a boolean/toggle)

src/components/Proposal/ProfileProposalItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+function ProfileProposalItem({ votedProposal }: Props) {
+  const t = useFormatMessage()
+  const { proposal } = votedProposal
+  const choiceIdx = votedProposal.choice - 1
+  const { choices, finish_at, scores } = proposal
+  const dateText = t(`page.home.open_proposals.${Time().isBefore(Time(finish_at)) ? 'ends_date' : 'ended_date'}`, {
+    value: Time(finish_at).fromNow(),
+  })
+  const maxScore = Math.max(...scores)
+  const resultMatches = maxScore === scores[choiceIdx]
```
1emu, line 32: I don't get this  👀 
You are checking if the vp of the voted choice is the same as the winning score?
Could you find the winner choice and match that against the selected choice instead? 
(It makes more sense to me to compare by choices instead of by score)

src/components/Proposal/ResultStance.tsx
```diff
@@ -0,0 +1,29 @@
...
+import Cancel from '../Icon/Cancel'
+import CheckCircle from '../Icon/CheckCircle'
+
+import './ResultStance.css'
+
+interface Props {
+  resultMatches: boolean
+}
+
+function ResultStance({ resultMatches }: Props) {
```
1emu, line 15: Does this show while the proposal is open, or only on finished proposals?

I'm thinking if we should call this `result` or `outcome` (outcome would imply the proposal is already finished)

src/entities/Votes/types.ts
```diff
@@ -21,6 +21,9 @@ export type VotedProposal = SnapshotVote & {
     proposal_id: string
     type: ProposalType
     status: ProposalStatus
+    author: string
+    finish_at: number
+    scores: number[]
```
1emu, line 26: I'm wondering if we really need this `scores`, because the `SnapshotVote` already has the proposal with the scores inside it.

This part of the VotedProposal should only add that info that we don't store in snapshot.

You could also ask for the `author` in the same request to graphql.

1emu (changes requested):
Looking goood, left some comments!

The one about the snapshot vote id needs to be addressed before merging :)

src/intl/en.json
ncomerci: No helper text according to Gino. Deleted

src/entities/Votes/types.ts
ncomerci: By this I am just saying that the properties are not optional, they are included

image

1emu (approved):
🚢

andyesp (approved):

PR #586, opened 2022-10-25
feat: remove ipfs fetch on proposal creation

andyesp:
TODO: 
- [x] Remove snapshot_signature from proposal model.


1emu (approved):
😍 🚀 🚀 🚀 🚀

src/migrations/1666964653220_remove-snapshot-signature-column-in-proposal.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.dropColumn(Model.tableName, 'snapshot_signature')
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.addColumn(Model.tableName, {
+    snapshot_signature: {
+      type: 'TEXT',
+      notNull: true,
```
ncomerci, line 16: if it is not null, shouldn't be a default value?

ncomerci (approved):

src/migrations/1666964653220_remove-snapshot-signature-column-in-proposal.ts
andyesp: Use the same object as in the create proposals migration https://github.com/decentraland/governance/blob/master/src/migrations/1616763227514_create-proposals.ts#L26.
I think if it is notNull without default value it means you can't create a row without that value. If it has a default value you can avoid passing a value to the row and the default is set (and it is always not null).

PR #589, opened 2022-10-28
feat: add option to refresh votes in votes endpoint

1emu:


src/entities/Votes/routes.ts
```diff
@@ -14,19 +14,21 @@ import { Vote, VoteAttributes } from './types'
 import { createVotes, toProposalIds } from './utils'
 
 export default routes((route) => {
-  route.get('/proposals/:proposal/votes', handleAPI(getProposalVotes))
+  route.get('/proposals/:proposal/votes?:refresh', handleAPI(getProposalVotes))
```
andyesp, line 17: I think query params are not necessary when defining the route path 🤔

andyesp (approved):

PR #595, opened 2022-11-02
feat: fetch all comments from discourse

1emu:


src/modules/discourse/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+        const newComments = newPostsResponse.post_stream.posts
+        allComments = [...allComments, ...newComments]
+        if (newComments.length < DISCOURSE_BATCH_SIZE) {
+          hasNext = false
+        } else {
+          skip = allComments.length
+        }
+      }
+    } catch (error) {
+      console.error(`Error while fetching discourse posts in batches: `, error)
```
1emu, line 23: maybe send to rollbar? (how would we do that?)

src/modules/discourse/utils.ts
```diff
@@ -0,0 +1,27 @@
+import { Discourse, DiscoursePostInTopic } from '../../clients/Discourse'
+
+export async function fetchAllComments(discourseTopicId: number) {
```
andyesp, line 3: I think this can be part of DiscourseService 🤔

andyesp (approved):

src/modules/discourse/utils.ts
1emu: I tried, but the service is not avail to the front :/

andyesp (approved):

PR #596, opened 2022-11-03
feat: sentiment voting

1emu:
- [x] configurable topics per proposal type
- [x] sentiment survey interface
- [x] survey encoder/decoder
- [x] show survey results
- [x] don't prompt survey on proposals without survey topics
- [x] 140 chars limit validation on reason
- [x] handle voting transaction rejected on metamask
- [x] handle unauthorized snapshot error on voting
- [x] move threshold data
- [x] parcial/total results component 
- [x] internationalization
- [x] Added final topics to all grants
- [x] Added validation for incoming surveys with unavailable topics
- [x] Updated snapshot version to use metadata instead of reason
- [x]  Updated encoder/decoder to encode a string object and decode an object
- [x]  Decoder only accepts surveys with correct topic format and valid reasons
- [x]  Removed copy feedback option
- [x]  Removed topic label field, and made topic_id unique without a char limit
- [x]  Added foreign key constraint to proposal survey topic table

image

image

image

image

Outside of scope:
- [ ] floating magic bar
- [ ] Votes detailed results with sentiment feedback


src/components/Proposal/SentimentSurvey/SentimentSurvey.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { useEffect } from 'react'
+
+import { Loader } from 'decentraland-ui'
```
andyesp, line 3: This is probably causing window not defined error

src/components/Common/Pipe.css
```diff
@@ -0,0 +1,10 @@
+.Pipe {
+  width: 20px;
+  height: 0;
+  left: 755px;
```
andyesp, line 4: Are these left and top styles doing anything? Looks weird without an absolute positioning and without context

src/clients/SnapshotApi.ts
```diff
@@ -139,15 +139,18 @@ export class SnapshotApi {
...
   ): Promise<SnapshotReceipt> {
     const voteMessage = {
       space: SnapshotApi.getSpaceName(),
       proposal: proposalSnapshotId,
       type: SNAPSHOT_PROPOSAL_TYPE,
       choice: choiceNumber,
+      reason: comment,
       app: SNAPSHOT_APP_NAME,
     }
+    throw new Error('testeando error')
```
andyesp, line 153: 🤗

src/clients/SnapshotApi.ts
```diff
@@ -139,13 +139,15 @@ export class SnapshotApi {
     account: Web3Provider | Wallet,
     address: string,
     proposalSnapshotId: string,
-    choiceNumber: number
+    choiceNumber: number,
+    comment?: string
```
andyesp, line 143: Why don't we call this `reason`?

src/components/Icon/AngryEmoji.tsx
```diff
@@ -0,0 +1,34 @@
...
+      viewBox="0 0 23 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path d="M0 23.61H23V0.61H0V23.61Z" fill="url(#pattern2)" />
+      <defs>
+        <pattern id="pattern2" patternContentUnits="objectBoundingBox" width="1" height="1">
+          <use xlinkHref="#image0_7148_2825" transform="scale(0.00625)" />
+        </pattern>
+        <image
```
andyesp, line 23: As this use an image... does it make sense to use an svg? Can't we use a real emoji?

src/components/Icon/GreyX.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+function GreyX({ className }: { className?: string }) {
```
andyesp, line 3: Maybe `Cross`? Do we need to specify the color here?

src/components/Icon/PartyEmoji.tsx
```diff
@@ -0,0 +1,34 @@
...
...Y4PLBxfX7J7MtnXABtmdiVLGSbYyBkNTLt19IgsKlNOVGWelnTJAGS48hgR0ddim/9uyubWBqnPKhKr/x8NBMAgoL20/thgavk9Xbnt3yLFnXaz9IiGpNEss1jrWSysWuj14Vd2dhf+3HbXTgoLhbxvW6J/qrUFpzDf+KrOoyjVze2r/fb1cT/e1x929nbXk2ttmu7KNbblmd4ErZuAgYCBgIUYS14o2W+pdFl2kwsslFvvA6RzFl1Isdhq107HUI9HKn9k17w4lEE/snM2Pt5u5OuzjdyNhgV8yIwb3E/eu+eT7zy7rbU3oiyFiyelcNV5MgKnRJzVJbIXbzt4diZap8oiU09r4AW6ZAAGIgARiW8Z8LcKf9kTF7kXvhmAzWfOfvXEppmPrqvmhyLFqLsyZmw0qD8wWFhjdFfTh1ur63de1Vv63PQwHkL6We1c4IiQJ6xEopXqHW9vxnHMiW+AM6QZt5hFp59vmuv2zU6dJtvZitle1tk8iO0ma/JptrbFxjaYOWLmkBasGZSBKLZC9TXTWiTild2zvIqaWkoMn2yp+PTuucG6IeorTk276WtiOCtQwpBEWkaFAaAlh3jj3i8/AMIDF+wjA8Ccw4X1NKYnDMCnnAhQOgcxf2yY8K09YV5J0xJaM62sGZxdtT1aG3z2qtUzn5gbdo8QA1DFNHqLMmDhyPoaDxY+baZn6wIxCHEj6p6da+OA8wEqgIfIDIGZBTMiMEswSy69gBZudr0GUW7ZQnKKRt1gBBIObAJABE1Vfw1wrhhtXi3vNXJ+0rHLBniBLg8AMmCVxPq2edgoiqHte1bX7I3rXbM7yRjDbn7P1tXFv9o+WPlmXWvvOxqvqzHyTdW/7b8JXPRH8Eg7Bxz1uRauIGTnUPSDp1AWv9dE40kKKPxzVXNlQ2sm/eZSjl82dBkBUGBt2zxMpEDaHIyH9r3rXfM2Q+JfOpx/dld/MRWGizIymDw4k85tdP2UXuTqFaGY16U0/Mk4gn9hQVelywOAAMCAMAaGXZOJ8EUh8CXDGBARtJCoWe2vLGgcMOODdz5rcRIwn2DLn/R1v1Cgq9LlA8Ax8pGLwfmiYjg/GMfPn+u6n/Xg/ywA+QtDly0AnwCdj/tdzHWXCohL/d0vBeDG6ZcJgON0rgH/WXHAXwod7snS/wHy8qGzhjh4nQAAAABJRU5ErkJggg=="
```
andyesp, line 27: Same comment as in `AngryEmoji` icon

src/components/Icon/PokerFaceEmoji.tsx
```diff
@@ -0,0 +1,35 @@
...
+      viewBox="0 0 23 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path d="M0 23.61H23V0.61H0V23.61Z" fill="url(#pattern1)" />
+      <defs>
+        <pattern id="pattern1" patternContentUnits="objectBoundingBox" width="1" height="1">
+          <use xlinkHref="#image0_7148_2824" transform="scale(0.00625)" />
+        </pattern>
+        <image
```
andyesp, line 24: Same comment as in `AngryEmoji`

src/components/Proposal/ProposalComments.tsx
```diff
@@ -16,12 +16,12 @@ import './ProposalComments.css'
 
 const DEFAULT_SHOWN_COMMENTS = 5
 
-export type ProposalResultSectionProps = Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> & {
+export type ProposalGovernanceSectionProps = Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> & {
```
andyesp, line 19: I think this should be `ProposalComents` 🤔

src/components/Proposal/SentimentSurvey/SentimentSurveyRow.css
```diff
@@ -0,0 +1,68 @@
+.SentimentSurveyRow {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  border-top: 1px solid rgba(115, 110, 125, 0.1);
```
andyesp, line 5: This color can be in the theme.

src/components/Proposal/SentimentSurvey/SurveyTopicResult.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div className="SurveyTopicResult">
+      <span className="SurveyTopicResult__Header">{t(`survey.survey_topics.${topicLabel}`)}</span>
+      <div className="SurveyTopicResult__ReactionContainer">
+        {reactions.map((reactionType, index) => (
+          <ReactionCounter
+            key={`ReactionCounter__${index}`}
+            reactionType={reactionType as ReactionType}
+            count={topicResult[reactionType as ReactionType]}
+            drawPipe={index < reactions.length - 1}
```
andyesp, line 27: If this is showing only on the last component. Can't we do something like...?
```
<div>
  {reactions.map((reactionType, index) => (
    <ReactionCounter />
  }
  <Pipe />
}
```

src/components/Section/ProposalGovernanceSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Vote } from '../../entities/Votes/types'
+import { ProposalPageOptions, SelectedChoice } from '../../pages/proposal'
+import { ChoiceProgressProps } from '../Status/ChoiceProgress'
+
+import ProposalVotingSection from './ProposalVoting/ProposalVotingSection'
+
+import './DetailsSection.css'
+import VotingStatusSummary from './VotingStatusSummary'
+
+export type ProposalGovernanceSectionProps = Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> & {
```
andyesp, line 16: Does this need to be exported?

src/components/Section/ProposalResults.tsx
```diff
@@ -0,0 +1,78 @@
...
+import Divider from '../Common/Divider'
+import Lock from '../Icon/Lock'
+import ChoiceProgress, { ChoiceProgressProps } from '../Status/ChoiceProgress'
+
+import './DetailsSection.css'
+import { ProposalPromotionSection } from './ProposalPromotionSection'
+import './ProposalResults.css'
+import VotingStatusSummary from './VotingStatusSummary'
+
+export type ProposalGovernanceSectionProps = Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> & {
```
andyesp, line 23: This should be called `ProposalResults`. Also don't need the export I think.

src/components/Section/ProposalResults.tsx
```diff
@@ -0,0 +1,78 @@
...
+  proposal?: ProposalAttributes | null
+  userHasVoted?: boolean
```
andyesp, line 25: This prop isn't used.

src/components/Section/ProposalResults.tsx
andyesp: Don't need to extend from HTMLDivElement either.

src/components/Section/ProposalVoting/ChoiceButtons.tsx
```diff
@@ -28,24 +36,47 @@ export const ChoiceButtons = ({
...
   onVote,
+  selectedChoice,
+  castingVote,
+  patchOptions,
   startAt,
+  showError,
+  onRetry,
 }: Props) => {
   const t = useFormatMessage()
+  const [timeLeft, setTimeLeft] = useState(SECONDS_FOR_VOTING_RETRY)
```
andyesp, line 47: This logic can be extracted into a hook to reuse it.

src/components/Section/ProposalVoting/ChoiceButtons.tsx
```diff
@@ -28,24 +36,47 @@ export const ChoiceButtons = ({
...
+        setTimeLeft((timeLeft) => timeLeft - 1)
+      }, 1000)
+
+      if (timeLeft === 0) {
+        setTimeLeft(SECONDS_FOR_VOTING_RETRY)
+        onRetry()
+      }
+      return () => clearTimeout(timer)
+    }
+  }, [showError, timeLeft, onRetry])
```
andyesp, line 61: This logic can be extracted into a hook to reuse it.

src/components/Section/ProposalVoting/ChoiceButtons.tsx
```diff
@@ -28,24 +36,47 @@ export const ChoiceButtons = ({
...
         const votedCurrentChoice = vote?.choice === currentChoiceIndex + 1
         const delegateVotedCurrentChoice = delegateVote?.choice === currentChoiceIndex + 1
         return (
           <ChoiceButton
+            selected={selectedChoice.choice === currentChoice}
             key={currentChoice}
             voted={votedCurrentChoice}
             disabled={votedCurrentChoice || !started}
-            onClick={(e: React.MouseEvent<unknown>) => onVote && onVote(e, currentChoice, currentChoiceIndex + 1)}
+            onClick={() => {
```
andyesp, line 77: I'd extract this logic to a `handleChoiceClick` function to simplify reading.

src/components/Proposal/SentimentSurvey/SentimentSurvey.tsx
```diff
@@ -0,0 +1,70 @@
...
+import SentimentSurveyRow from './SentimentSurveyRow'
+
+interface Props {
+  survey: Survey
+  surveyTopics: Pick<SurveyTopicAttributes, 'label' | 'topic_id'>[] | null
+  isLoadingSurveyTopics: boolean
+  setSurvey: React.Dispatch<React.SetStateAction<Survey>>
+}
+
+const SentimentSurvey = ({ survey, surveyTopics, isLoadingSurveyTopics, setSurvey }: Props) => {
```
ncomerci, line 17: Did you define the function as a variable for any special reason? I know it doesn't affect anything but it caught my attention 🤔

src/components/Proposal/SentimentSurvey/SurveyTopicResult.css
```diff
@@ -0,0 +1,54 @@
+.SurveyTopicResult {
+  height: 48px;
+  background: var(--white-900);
+  border: 1px solid rgba(115, 110, 125, 0.24);
```
ncomerci, line 4: idem colors

src/components/Section/ProposalVoting/ProposalVotingSection.tsx
```diff
@@ -92,35 +106,50 @@ const ProposalVotingSection = ({ proposal, votes, loading, changingVote, choices
...
-              delegate={delegate}
-              delegateVote={delegateVote}
-              totalVotes={totalVotes}
-              onVote={onVote}
-              startAt={proposal?.start_at}
-            />
+          {!proposalContext.showSnapshotRedirect && (
+            <>
+              <Header sub>{'Get Involved'}</Header>
+              {/*TODO: internationalization*/}
```
ncomerci, line 112: Just a comment to not forget this

src/components/Common/SnapshotRedirectButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <Button
+      fluid
+      inverted={inverted}
+      primary={!inverted}
+      href={snapshotProposalUrl(proposal)}
+      target="_blank"
+      rel="noopener noreferrer"
+      className="SnapshotRedirectButton"
+    >
+      <div className="SnapshotRedirectButton__Hidden" />
```
andyesp, line 27: What's the purpose of this div?

src/clients/SnapshotApi.ts
1emu: image

src/components/Common/SnapshotRedirectButton.tsx
1emu: I wasn't able to put the label at the center with the external link icon to the right using absolute positioning, because when hovering over the button, the transformY() changed everything and the icon changed it's position.

The easy way out was using 3 elements, hide one, and let flex do the magic 🪄

src/components/Helper/HelperText.css
```diff
@@ -17,3 +17,7 @@
 .HelperText__Content > p.dg.Paragraph > a.dg.Link.Link--pointer {
   color: var(--blue-600);
 }
+
+.ui.bottom.center.popup.transition.visible.Helper__Container--Popup {
+  z-index: 3000;
```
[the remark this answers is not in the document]
1emu: yeah, the popup was not visible over the modal

src/components/Proposal/SentimentSurvey/SentimentSurvey.tsx
1emu: No special reason. But we do it in several places, so it's not a special case either

src/components/Common/SnapshotRedirectButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+      {t('component.snapshot_redirect_button_label')}
+      <OpenExternalLink
+        className="SnapshotRedirect__ButtonIcon"
```
andyesp, line 32: `SnapshotRedirectButton__Icon` 🙈

src/components/Helper/IconHelper.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Popup, PopupProps } from 'decentraland-ui/dist/components/Popup/Popup'
+
+type Props = Pick<PopupProps, 'position'> & {
+  text: string
+  containerClassName?: string
+  icon: string
+}
+
+function IconHelper({ position, text, containerClassName, icon }: Props) {
+  const useContainerClassName = containerClassName && containerClassName.length > 0
```
andyesp, line 14: I'd change this to `hasContainerClassName` as the `use*` prefix is used for hooks 🤔

src/components/Modal/Votes/VotingModal/VotingModalSurvey.tsx
```diff
@@ -0,0 +1,82 @@
...
+  surveyTopics,
+  isLoadingSurveyTopics,
+  onCastVote,
+  castingVote,
+  proposalContext,
+}: VotingModalSurveyProps) {
+  const t = useFormatMessage()
+  const { selectedChoice, showVotingError, retryTimer } = proposalContext
+
+  if (!selectedChoice.choiceIndex || !selectedChoice.choice) {
```
andyesp, line 40: I think you can move this check to `VotingModal`

`const showVotingModalSurvey = selectedChoice.choiceIndex && selectedChoice.choice`

src/components/Common/SnapshotRedirectButton.css
```diff
@@ -0,0 +1,13 @@
+.SnapshotRedirectButton {
+  display: flex !important;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 24px;
+  border: 1px solid rgba(115, 110, 125, 0.32) !important;
```
ncomerci, line 6: Can we use --black-100 here?

src/components/Modal/Votes/VotingModal/VotingModalSurvey.tsx
1emu: Ty! Yeah, the check is there but as I am passing the properties to the ModalSurvey inside the proposal context object, the validation isn't taken into account. 

I think we can do `formatChoice(selectedChoice.choice!)` for this one

src/components/Proposal/SentimentSurvey/SurveyTopicResult.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <div className="SurveyTopicResult">
+      <span className="SurveyTopicResult__Header">{t(`survey.survey_topics.${topicLabel}`)}</span>
+      <div className="SurveyTopicResult__ReactionContainer">
+        {reactions.map((reactionType, index) => (
+          <Fragment key={`ReactionCounter__${index}`}>
+            <ReactionCounter
+              reactionType={reactionType as ReactionType}
+              count={topicResult[reactionType as ReactionType]}
+            />
+            {index < reactions.length - 1 && <Pipe />}
```
ncomerci, line 29: I think this condition could be a constant with an informative name

src/components/Section/VotingResultsInfo.tsx
```diff
@@ -0,0 +1,61 @@
...
+              {timeout.time > 0
+                ? t('page.proposal_detail.time_left_label', { countdown: endDate.fromNow() })
+                : t('page.proposal_detail.voting_section.voting_finished', { timeElapsed: endDate.fromNow() })}
+            </Bold>
+          </div>
+        </div>
+      )}
+      {timeout.time <= 0 && (
+        <div className="VotingResultsInfo__Container">
+          <div className="VotingResultsInfo__Subtitle">{}</div>
```
ncomerci, line 47: why is this empty? 🤔

src/components/Section/VotingResultsInfo.tsx
```diff
@@ -0,0 +1,61 @@
...
+          <div className="VotingResultsInfo__Title">
+            <Bold>{t('page.proposal_detail.voting_section.voting_finished', { timeElapsed: endDate.fromNow() })}</Bold>
+          </div>
+        </div>
+      )}
+      <div className="VotingResultsInfo__Container">
+        <div className="VotingResultsInfo__Subtitle">{'Voting activity'}</div>
```
ncomerci, line 54: i18n

src/components/Common/SnapshotRedirectButton.css
1emu: The thing is that it needs to be the same as the choice buttons border, which is slightly lighter. I'll ask Juani if we can add it to the theme or change both.

src/components/Proposal/SentimentSurvey/ReactionCounter.tsx
```diff
@@ -0,0 +1,36 @@
...
+const ReactionCounter = ({ reactionType, count }: Props) => {
+  const t = useFormatMessage()
+  const reactionView = REACTIONS_VIEW.find((reactionIcon) => reactionIcon.reaction === reactionType)
+  if (!reactionView) {
+    console.log(`Missing icon for reaction type ${reactionType}`)
+    return null
+  }
+
+  return (
+    <>
```
andyesp, line 23: This fragment is not necessary

src/components/Proposal/SentimentSurvey/SentimentSurvey.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const t = useFormatMessage()
+
+  useEffect(() => {
+    if (!isLoadingSurveyTopics) {
```
andyesp, line 22: I think you can group these two conditions
`if (!isLoadingSurveyTopics && survey.length === 0) {`

src/components/Proposal/SentimentSurvey/SentimentSurveyRow.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (reaction && reaction !== ReactionType.EMPTY) {
+      setShowReactions(false)
+      setPickedReaction(reaction)
+    }
+  }, [reaction])
+
+  return (
+    <div
+      className={TokenList.join(['SentimentSurveyRow', showAddReaction && 'SentimentSurveyRow__Expanded'])}
+      onMouseEnter={() => !reactionPicked && setShowAddReaction(true)}
```
andyesp, line 54: You can avoid this logic using css and the :hover pseudo-class in SentimentSurveyRow and changing the opacity in SentimentSurveyRow__AddReactionLabel.

src/pages/proposal.tsx
```diff
@@ -57,59 +65,112 @@ import './proposals.css'
...
+export type SelectedChoice = { choice?: string; choiceIndex?: number }
+const EMPTY_CHOICE_SELECTION: SelectedChoice = { choice: undefined, choiceIndex: undefined }
 const PROPOSAL_STATUS_WITH_UPDATES = new Set([ProposalStatus.Passed, ProposalStatus.Enacted])
+const EMPTY_CHOICES: string[] = []
+const MAX_ERRORS_BEFORE_SNAPSHOT_REDIRECT = 3
+const SECONDS_FOR_VOTING_RETRY = 5
 
-type ProposalPageOptions = {
-  changing: boolean
+export type ProposalPageContext = {
```
andyesp, line 75: I'm not sure about naming this Context as Context in a react app _context_ is a core react API feature... 🤔

andyesp: I think I prefer state as that's what this is. (`ProposalPageState`)

src/pages/proposal.tsx
```diff
@@ -57,59 +65,112 @@ import './proposals.css'
...
     [account, subscriptions]
   )
-  const [voting, vote] = useAsyncTask(
-    async (_: string, choiceIndex: number) => {
-      if (proposal && account && provider && votes) {
+
+  const proposalResults = useRef<HTMLDivElement>(null)
+
+  const handleScrollClick = () => {
+    proposalResults.current?.scrollIntoView({ behavior: 'smooth' })
```
andyesp, line 133: It looks like options param in `scrollIntoView` are not supported on Safari. https://developer.mozilla.org/es/docs/Web/API/Element/scrollIntoView
Is that right? Are we ok with that?

src/pages/proposal.tsx
```diff
@@ -224,6 +298,20 @@ export default function ProposalPage() {
   const showProposalUpdates = publicUpdates && isProposalStatusWithUpdates && proposal?.type === ProposalType.Grant
   const showImagesPreview =
     !proposalState.loading && proposal?.type === ProposalType.LinkedWearables && !!proposal.configuration.image_previews
+  const showSurvey = !isLoadingSurveyTopics && surveyTopics && surveyTopics.length > 0
+
+  const getVotingMethod = () => {
```
andyesp, line 303: I'd change this to
```
const handleVoteClick = (selectedChoice: SelectedChoice) => {
    if (showSurvey) {
      updateContext({
        selectedChoice: selectedChoice,
        showVotingModal: true,
      })
    } else {
      castVote(selectedChoice)
    }
  }
```

src/pages/proposal.tsx
1emu: Ty, just tested this on Safari and it works :)

src/entities/ProposalSurveyTopics/routes.ts
```diff
@@ -0,0 +1,8 @@
+import handleAPI from 'decentraland-gatsby/dist/entities/Route/handle'
+import routes from 'decentraland-gatsby/dist/entities/Route/routes'
+
+import { getProposalSurveyTopics } from './utils'
+
+export default routes((route) => {
+  route.get('/proposals/:proposal/survey-topics', handleAPI(getProposalSurveyTopics))
```
andyesp, line 7: This one should be a function in a service not in a utils file 🤔

src/components/Modal/Votes/VotingModal/VotingModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return null
+  }
+
+  return (
+    <Modal
+      size="tiny"
+      className="VotingModal ProposalModal"
+      open={proposalPageState.showVotingModal}
+      closeIcon={<Close />}
+      onClose={() => {
```
andyesp, line 47: I think you can write this `onClose={onClose}`

src/components/Modal/Votes/VotingModal/VotingModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+export function VotingModal({
+  proposal,
+  onClose,
+  surveyTopics,
+  isLoadingSurveyTopics,
+  onCastVote,
+  castingVote,
+  proposalPageState,
+}: VotingModalProps) {
+  const [survey, setSurvey] = useState<Survey>([])
```
andyesp, line 34: If this is only used in VotingModalSurvey then you can move this to that component.

src/components/Proposal/ProposalActions.tsx
```diff
@@ -0,0 +1,86 @@
...
+          basic
+          fluid
+          loading={deleting}
+          disabled={proposal?.status !== ProposalStatus.Pending && proposal?.status !== ProposalStatus.Active}
+          onClick={() => updatePageState({ confirmDeletion: true })}
+        >
+          {t('page.proposal_detail.delete')}
+        </Button>
+      )}
+      {isCommittee && (proposal?.status === ProposalStatus.Passed || proposal?.status === ProposalStatus.Enacted) && (
```
andyesp, line 42: I'd move all these conditions to some boolean variables so they are easier to read and understand what is being rendered.
`showDeleteButton`, `showEnactButton`, `showPassButton`

src/pages/proposal.tsx
```diff
@@ -57,59 +65,112 @@ import './proposals.css'
 
 // TODO: Review why proposals.css is being imported and use only proposal.css
 
+export type SelectedChoice = { choice?: string; choiceIndex?: number }
```
andyesp, line 68: Can you move this to `Proposal/types.ts`? or maybe `Votes/types.ts`? Also it may be useful to rename it to `SelectedVoteChoice`. A topic feedback can be a choice in this context too 🤔

src/pages/proposal.tsx
```diff
@@ -57,59 +65,112 @@ import './proposals.css'
...
+const EMPTY_CHOICE_SELECTION: SelectedChoice = { choice: undefined, choiceIndex: undefined }
 const PROPOSAL_STATUS_WITH_UPDATES = new Set([ProposalStatus.Passed, ProposalStatus.Enacted])
+const EMPTY_CHOICES: string[] = []
```
andyesp, line 71: `EMPTY_VOTE_CHOICES`?

src/pages/proposal.tsx
```diff
@@ -252,138 +338,131 @@ export default function ProposalPage() {
...
                   proposal={proposal}
                   votes={votes}
-                  changingVote={options.changing}
-                  onChangeVote={(_, changing) => patchOptions({ changing })}
-                  onOpenVotesList={() => patchOptions({ showVotesList: true })}
-                  onVote={(_, choice, choiceIndex) => vote(choice, choiceIndex)}
+                  partialResults={partialResults}
+                  choices={choices}
+                  castingVote={castingVote}
+                  onChangeVote={(_, changing) => updatePageState({ changingVote: changing })}
```
andyesp, line 381: If we are using the handler pattern in all other function props here I'd move this to a `handleChangeVote` function.

src/pages/proposal.tsx
```diff
@@ -57,59 +65,112 @@ import './proposals.css'
...
+          )
+          updatePageState({
+            changingVote: false,
+            showVotingModal: false,
+            showVotingError: false,
+            confirmSubscription: !votes![account!],
+          })
+          votesState.reload()
+        } catch (err) {
+          console.error(err, { ...(err as Error) }) //TODO report error
```
andyesp, line 157: If we are leaving this todo please write it like `// TODO: report error` so it is findable in a quick search 😃

src/entities/ProposalSurveyTopics/utils.ts
```diff
@@ -0,0 +1,15 @@
+import RequestError from 'decentraland-gatsby/dist/entities/Route/error'
+import { Request } from 'express'
+
+import ProposalModel from '../Proposal/model'
+import SurveyTopicModel from '../SurveyTopic/model'
+
+export async function getProposalSurveyTopics(req: Request<{ proposal: string }>) {
+  try {
+    const id = req.params.proposal
+    const proposal = await ProposalModel.getProposal(id)
```
andyesp, line 10: If we pass the category instead of the proposal we can avoid the getProposal call here.

src/components/Section/VotingResultsInfo.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const timeout = useCountdown(endDate)
+
+  const voteCount = useMemo(
+    () => partialResults.reduce((sum, choiceProgress) => sum + choiceProgress.votes, 0),
+    [partialResults]
+  )
+
+  return (
+    <div className="VotingResultsInfo">
+      {timeout.time > 0 && (
```
andyesp, line 31: You can move this to a variable... `isCountdownRunning` or something?

src/components/Section/VotingResultsInfo.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <div className="VotingResultsInfo__Title">
+            <Bold>
+              {timeout.time > 0
+                ? t('page.proposal_detail.time_left_label', { countdown: endDate.fromNow() })
+                : t('page.proposal_detail.voting_section.voting_finished', { timeElapsed: endDate.fromNow() })}
+            </Bold>
+          </div>
+        </div>
+      )}
+      {timeout.time <= 0 && (
```
andyesp, line 45: And `!isCountdownRunning`?

src/entities/SurveyTopic/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { ReactionType, Survey } from './types'
+
+type ReactionView = { reaction: ReactionType; icon: string }
+export const REACTIONS_VIEW: ReactionView[] = [
```
andyesp, line 4: Why is it called "view"? Maybe `REACTIONS_LIST`?

src/entities/SurveyTopic/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { ReactionType, Survey } from './types'
+
+type ReactionView = { reaction: ReactionType; icon: string }
```
andyesp, line 3: Type instead of reaction as this is a list of reactions? 
`{ type: ReactionType; icon: string }`

src/intl/en.json
```diff
@@ -265,6 +265,32 @@
...
+  "survey": {
+    "topics_title": "Topics",
+    "reactions_title": "Reaction",
+    "survey_topics": {
+      "idea": "Idea",
+      "personnel": "Personnel",
+      "budget": "Budget",
+      "roadmap": "Roadmap Proposal",
+      "potential_impact": "Potential Impact",
+      "test_topic_1": "Test Topic 1",
```
andyesp, line 277: Test topics here 🙈

src/clients/Governance.ts
```diff
@@ -353,4 +354,13 @@ export class Governance extends API {
 
     return response.data
   }
+
+  async getSurveyTopics(proposalId: string) {
+    const result = await this.fetch<ApiResponse<Topic[]>>(
+      `/proposals/${proposalId}/survey-topics`,
```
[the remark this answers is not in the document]
1emu: The thing is you need the type, and if the type is a grant, you also need the category.

I think for this case it's better if we request the survey topics for a proposal only by the `id`, without needing to know what specific fields the backend needs to answer the request.
This way, if we ever need to change the implementation, this part wouldn't need to change.

WDYT?

src/entities/ProposalSurveyTopics/type.ts
```diff
@@ -0,0 +1,10 @@
+import { ProposalType } from '../Proposal/types'
+
+export type ProposalSurveyTopicAttributes = {
+  id: string
+  topic_id: string
+  proposal_type: ProposalType
+  proposal_sub_types: string | null
```
andyesp, line 7: We could type this instead of using `string` 🤔

src/components/Modal/Votes/VotingModal/VotingModal.tsx
1emu: Right, I forgot to move this when I removed the `copy survey` part.

src/entities/ProposalSurveyTopics/type.ts
```diff
@@ -0,0 +1,10 @@
+import { ProposalType } from '../Proposal/types'
+
```
andyesp, line 2: Use `types.ts` in plural to name this file instead of `type.ts` 🙏

src/components/Proposal/SentimentSurvey/SentimentSurveyRow.css
```diff
@@ -0,0 +1,75 @@
...
+  color: var(--black-600);
+  position: relative;
+  opacity: 0;
+}
+
+.SentimentSurveyRow:hover .SentimentSurveyRow__AddReactionLabel {
+  opacity: 1 !important;
+}
+
+#slide {
```
andyesp, line 42: Maybe I'm missing something but can't you use `SentimentSurveyRow__AddReaction`?

src/components/Proposal/SentimentSurvey/SentimentSurveyRow.tsx
```diff
@@ -0,0 +1,99 @@
...
+  reaction?: ReactionType | null
+  onReactionPicked: (topic: Topic, reaction: ReactionType) => void
+  onReactionUnpicked: (topic: Topic) => void
+}
+
+const SentimentSurveyRow = ({ topic, reaction, onReactionPicked, onReactionUnpicked }: Props) => {
+  const t = useFormatMessage()
+  const [showReactions, setShowReactions] = useState(false)
+  const [pickedReaction, setPickedReaction] = useState<ReactionType | null>()
+  const reactionPicked = pickedReaction != null
```
andyesp, line 24: `hasPickedReaction` 😄

src/entities/SurveyTopic/types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export type SurveyTopicAttributes = {
+  id: string
+  topic_id: string
+  created_at: Date
+  updated_at: Date
+}
+
+export type TopicFeedback = {
+  topic: Topic
```
andyesp, line 17: What if this is only `topic_id` instead of topic with an object?

```
type TopicFeedback = {
  topic_id: SurveyTopicAttributes['topic_id']
  reaction: ReactionType
}
```
🤔

src/pages/proposal.tsx
1emu: We are not jiji. I think that most functions that are just an update on the state are all anonymous functions, and if there are more steps then they are extracted to a handler.

Some examples:


```typescript
onOpenVotesList={() => updatePageState({ showVotesList: true })}
onClose={() => updatePageState({ showVotesList: false })}
onClose={() => updatePageState({ confirmSubscription: false })}
onClose={() => updatePageState({ confirmDeletion: false })}
```

I could extract them all to handlers though. WDYT? 
I don't mind them inline also, because it's getting long to read and declarations will be far up from usages.

andyesp: I don't mind. It's just weird to read handlers and inline functions in the same component props.

ncomerci (approved):

src/components/Section/ProposalGovernanceSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const showProposalThresholdsSummary = !!(
+    proposal &&
+    proposal?.required_to_pass != null &&
+    proposal?.required_to_pass >= 0 &&
+    !(proposal.status === ProposalStatus.Passed)
+  )
+
+  return (
+    <div
+      {...props}
```
andyesp, line 59: I don't think we need this spread here

src/components/Section/ProposalGovernanceSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { ProposalPageState, SelectedChoice } from '../../pages/proposal'
+import { ChoiceProgressProps } from '../Status/ChoiceProgress'
+
+import ProposalVotingSection from './ProposalVoting/ProposalVotingSection'
+
+import './DetailsSection.css'
+import { ProposalPromotionSection } from './ProposalPromotionSection'
+import ProposalThresholdsSummary from './ProposalThresholdsSummary'
+
+type ProposalGovernanceSectionProps = Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> & {
```
andyesp, line 17: You can skip this `Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> &` as we are not using specific div attributes in this component

src/components/Section/ProposalGovernanceSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <div
+      {...props}
+      className={TokenList.join([
+        'DetailsSection',
+        disabled && 'DetailsSection--disabled',
+        loading && 'DetailsSection--loading',
+        'ResultSection',
+        props.className,
```
andyesp, line 65: Not using className as a prop

src/components/Proposal/SentimentSurvey/SurveyResults.tsx
```diff
@@ -0,0 +1,93 @@
...
+    })
+  })
+  return topicsResults
+}
+
+const SurveyResults = ({ votes, isLoadingVotes, surveyTopics, isLoadingSurveyTopics }: Props) => {
+  const t = useFormatMessage()
+  const topicResults = useMemo(() => getResults(surveyTopics, votes), [surveyTopics, votes])
+  const topicIds = Object.keys(topicResults)
+  const thereAreVotes = votes && Object.keys(votes).length > 0 && !isLoadingVotes
```
andyesp, line 62: `hasVotes` 😁

src/components/Proposal/SentimentSurvey/SurveyResults.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const thereAreSurveyTopics = surveyTopics && surveyTopics?.length > 0 && !isLoadingSurveyTopics
```
andyesp, line 63: `hasSurveyTopics` 😁

src/components/Proposal/SentimentSurvey/SurveyResults.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <div className="SurveyResults">
+      <Divider />
+      {isLoadingSurveyTopics && <Loader />}
+      {!isLoadingSurveyTopics && (
+        <div>
+          <div className="SurveyResults__Header">
+            <Header>{t('survey.results.title')}</Header>
+          </div>
+          {topicIds.map((topicId: string, index: any) => {
```
andyesp, line 78: `index` is number

src/components/Proposal/SentimentSurvey/SurveyResults.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return availableTopics.find((topic) => topic.topic_id === topicFeedback.topic.topic_id)
+}
+
+function getResults(availableTopics: Topic[] | null, votes: Record<string, Vote> | null) {
+  if (!availableTopics || !votes) return {}
+  const topicsResults = initializeTopicResults(availableTopics)
+  Object.keys(votes).map((key) => {
+    const survey = decodeSurvey(votes[key].metadata)
+    survey.map((topicFeedback) => {
+      if (isTopicAvailable(availableTopics, topicFeedback) && topicFeedback.reaction != ReactionType.EMPTY) {
```
andyesp, line 50: I'd use `!==` unless there is a reason to use != 🤔

src/entities/SurveyTopic/types.ts
1emu: Isn't this the same?
`export type Topic = Pick<SurveyTopicAttributes, 'topic_id'>`

1emu: Oh, I can make it 

```typescript
export type TopicFeedback = Topic & {
  reaction: ReactionType
}
```

src/entities/SurveyTopic/utils.ts
1emu: Changed to

```typescript
type ReactionWithIcon = { reaction: Reaction; icon: string }
export const REACTION_LIST: ReactionWithIcon[] = [
  { reaction: Reaction.LOVE, icon: '🥰' },
  { reaction: Reaction.LIKE, icon: '😊' },
  { reaction: Reaction.NEUTRAL, icon: '😐' },
  { reaction: Reaction.CONCERNED, icon: '🤨' },
  { reaction: Reaction.EMPTY, icon: '-' },
]
```

src/components/Proposal/SentimentSurvey/SentimentSurveyRow.tsx
```diff
@@ -0,0 +1,99 @@
...
+  reaction?: Reaction | null
+  onReactionPicked: (topic: Topic, reaction: Reaction) => void
+  onReactionUnpicked: (topic: Topic) => void
+}
+
+const SentimentSurveyRow = ({ topic, reaction, onReactionPicked, onReactionUnpicked }: Props) => {
+  const t = useFormatMessage()
+  const [showReactions, setShowReactions] = useState(false)
+  const [pickedReaction, setPickedReaction] = useState<Reaction | null>()
+  const hasPickedReaction = pickedReaction != null
```
andyesp, line 24: You could probably change this to `!==` here too

src/components/Proposal/SentimentSurvey/SentimentSurveyRow.tsx
```diff
@@ -0,0 +1,99 @@
...
+        </div>
+      )}
+
+      {hasPickedReaction && (
+        <div className="SentimentSurveyRow__Reactions SentimentSurveyRow__PickedReaction">
+          {REACTION_LIST.map((reactionWithIcon, index) => {
+            if (reactionWithIcon.reaction !== Reaction.EMPTY) {
+              return (
+                pickedReaction === reactionWithIcon.reaction && (
+                  <div key={`Reaction__${index}`} onClick={() => changeReaction()}>
```
andyesp, line 86: You can write `onClick={changeReaction}`

ncomerci (approved):

src/components/Proposal/ProposalActions.tsx
```diff
@@ -0,0 +1,89 @@
...
+  updatePageState: (newState: Partial<ProposalPageState>) => void
+}
+
+export default function ProposalActions({ proposal, deleting, updatingStatus, updatePageState }: ProposalActionsProps) {
+  const t = useFormatMessage()
+  const [account] = useAuthContext()
+  const { isCommittee } = useIsCommittee(account)
+  const { isOwner } = useIsProposalOwner(proposal)
+
+  if (!proposal) return null
```
andyesp, line 25: As previously discussed 😛... I'd move this check to where ProposalActions is rendered.

andyesp: Good case here as proposal is needed in a hook before this condition...

src/components/Section/ProposalGovernanceSection.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const now = Time.utc()
+  const finishAt = Time.utc(proposal?.finish_at)
+  const finished = finishAt.isBefore(now)
+  const showPromotionSection = useMemo(
+    () => proposal && proposal.status === ProposalStatus.Passed && PROMOTABLE_PROPOSALS.includes(proposal.type),
+    [proposal]
+  )
+  const showProposalThresholdsSummary = !!(
+    proposal &&
+    proposal?.required_to_pass != null &&
```
andyesp, line 57: Use `!==`

src/entities/Votes/types.ts
```diff
@@ -29,3 +29,5 @@ export type VotedProposal = SnapshotVote & {
 }
 
 export type ChoiceColor = 'approve' | 'reject' | number
+
+export type SelectedVoteChoice = { choice?: string; choiceIndex?: number }
```
andyesp, line 33: Why choice and choiceIndex are optional? 🤔

src/components/Section/ProposalUpdatesActions.tsx
```diff
@@ -14,17 +20,40 @@ import './ProposalUpdatesActions.css'
...
   currentUpdate,
+  pendingUpdates,
+  proposal,
 }: ProposalUpdatesActionsProps) {
   const t = useFormatMessage()
   const hasSubmittedUpdate = !!currentUpdate?.completion_date
 
+  const { isOwner } = useIsProposalOwner(proposal)
+  const { isCoauthor } = useIsProposalCoAuthor(proposal)
+  const showProposalUpdatesActions =
```
andyesp, line 38: I think we can move this to the parent component.

src/components/Section/ProposalUpdatesActions.tsx
```diff
@@ -14,17 +20,40 @@ import './ProposalUpdatesActions.css'
...
+    isProposalStatusWithUpdates(proposal?.status) && proposal?.type === ProposalType.Grant && (isOwner || isCoauthor)
+
+  const onPostUpdateClick = useCallback(() => {
+    if (proposal === null) {
```
andyesp, line 42: If render condition is moved to parent component, we can remove null as part of the proposal type and remove this check.

src/entities/SurveyTopic/model.ts
```diff
@@ -0,0 +1,28 @@
...
+  static async getSurveyTopic(
+    proposalType: ProposalType,
+    proposalConfiguration: Pick<ProposalAttributes, 'configuration'>
+  ): Promise<Topic[]> {
+    const proposalCategory = getProposalCategory(proposalType, proposalConfiguration)
+    return await this.query(SQL`
+    SELECT s.topic_id
+    FROM ${table(SurveyTopicModel)} s
+        INNER JOIN ${table(ProposalSurveyTopicModel)} ps ON ps."topic_id" = s."topic_id"
+        WHERE ps."proposal_type" = ${proposalType}
```
andyesp, line 24: Are we validating this is a valid proposal type?

src/entities/Proposal/utils.ts
```diff
@@ -215,3 +216,7 @@ export function isProposalInCliffPeriod(grant: GrantAttributes) {
   const now = Time.utc()
   return !isOneTimePayment && Time.unix(grant.enacted_at).add(CLIFF_PERIOD_IN_DAYS, 'day').isAfter(now)
 }
+
+export function getProposalCategory(proposalType: ProposalType, proposalConfiguration: any): string | null {
+  return proposalType === ProposalType.Grant ? proposalConfiguration.category : null
```
andyesp, line 221: As this is used directly in the model. Can we validate the category we are returning is a valid category?

src/components/Section/ProposalGovernanceSection.tsx
```diff
@@ -0,0 +1,93 @@
...
+    proposal?.required_to_pass >= 0 &&
+    !(proposal.status === ProposalStatus.Passed)
+  )
+
+  if (!showPromotionSection && !showProposalThresholdsSummary && !!finished) return null
```
ncomerci, line 62: Same as before

src/hooks/useIsProposalCoAuthor.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { CoauthorStatus } from '../entities/Coauthor/types'
+import { GrantAttributes, ProposalAttributes } from '../entities/Proposal/types'
+import { isSameAddress } from '../entities/Snapshot/utils'
+
+import useCoAuthorsByProposal from './useCoAuthorsByProposal'
+
+export default function useIsProposalCoAuthor(proposal: ProposalAttributes | GrantAttributes | null) {
+  const [account] = useAuthContext()
+  const coauthorsByProposal = useCoAuthorsByProposal(proposal)
```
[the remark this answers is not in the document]
ncomerci: I think it makes a request every time

src/entities/Votes/types.ts
1emu: It's to be able to have an empty choice selection when you haven't voted

1emu:
```typescript
const [proposalPageState, updatePageState] = usePatchState<ProposalPageState>({
    changingVote: false,
    confirmSubscription: false,
    ...
    selectedChoice: EMPTY_VOTE_CHOICE_SELECTION,
  })
  ```

File: src/entities/SurveyTopic/model.ts (hunk 70 above)
Comment (1emu) [reply src/entities/SurveyTopic/model.ts:24]: Good question!

This method is only invoked with the `proposal.type` we have in our DB, so there shouldn't be a problem

```typescript
     const proposal = await ProposalModel.getProposal(id)
     return await SurveyTopicModel.getSurveyTopic(proposal.type, proposal.configuration)
```

src/components/Section/ProposalGovernanceSection.tsx
1emu: For this particular case I'd rather leave them here. 
Its not a simple check, and each of the individual variables are used to render child components

src/components/Proposal/View/ProposalGovernanceSection.tsx
```diff
@@ -81,14 +93,14 @@ function ProposalResultSection({
...
+              'ProposalGovernanceSection__Results',
+              !finished && 'ProposalGovernanceSection__Results--current',
             ])}
           >
-            {results.map((result) => {
+            {partialResults.map((result) => {
               return (
                 <ChoiceProgress
-                  onClick={onOpenVotesList}
+                  // onClick={onOpenVotesList} TODO
```
andyesp, line 103: Should we remove this?

src/modules/votes/utils.ts
```diff
@@ -97,3 +97,7 @@ export function getPartyVotes(
 
   return { votesByChoices, totalVotes }
 }
+
+export function formatChoice(choice: string) {
+  return choice.charAt(0).toUpperCase() + choice.slice(1)
```
andyesp, line 102: You can use lodash's `upperFirst` instead

andyesp (commented):
`forum.svg` file can be deleted as we don't use it anymore

src/components/Common/SnapshotRedirectButton.tsx
andyesp: Maybe `SnapshotRedirectButton__Spacing` or something space-related?

ncomerci (approved):

src/components/Proposal/View/ProposalGovernanceSection.tsx
1emu: I missed this, tx. Changed it for `onClick={() => updatePageState({ showVotesList: true })}`

andyesp (approved):

src/components/Proposal/View/ProposalGovernanceSection.tsx
```diff
@@ -50,10 +57,17 @@ function ProposalResultSection({
...
+  const showPromotionSection = useMemo(
+    () => proposal && proposal.status === ProposalStatus.Passed && PROMOTABLE_PROPOSALS.includes(proposal.type),
+    [proposal]
+  )
+
   useEffect(() => {
     setShowResults(hasVoted || finished || !userAddress)
   }, [hasVoted, finished, userAddress])
 
+  if (!showPromotionSection && !!finished) return null
```
andyesp, line 69: Why the promotion section is the condition to render the rest? 🤔

1emu: Good catch, this got outdated when I removed the other show results section.

src/components/Proposal/SentimentSurvey/SentimentSurvey.css
```diff
@@ -0,0 +1,24 @@
...
+  padding-right: 20px;
+  padding-left: 20px;
+  margin-bottom: 12px;
+}
+
+.SentimentSurveyRow {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  border-top: 1px solid rgba(115, 110, 125, 0.1);
```
ncomerci, line 22: --alpha-black-200

ncomerci (approved):

PR #602, opened 2022-11-08
feat: discord integration

andyesp:


src/entities/Updates/utils.ts
```diff
@@ -67,3 +69,11 @@ export const isBetweenLateThresholdDate = (dueDate?: Date) => {
 
   return Time().isBefore(newDueDate)
 }
+
+export function updateUrl(updateId: string, proposalId: string) {
```
andyesp, line 73: I prefer using verbs to name functions (getUpdateUrl). And I think we are doing it this way on the code, except some old functions.

src/entities/Profile/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { GOVERNANCE_API } from '../../constants'
+
+export function profileUrl(userAddress: string) {
```
andyesp, line 3: Same here as in the updateUrl comment.

src/entities/Proposal/jobs.ts
```diff
@@ -91,41 +108,49 @@ export async function finishProposal(context: JobContext) {
...
+  const acceptedProposals: ProposalWithWinnerChoice[] = []
+  const rejectedProposals: ProposalWithWinnerChoice[] = []
 
   for (const proposal of pendingProposals) {
     const outcome = await calculateOutcome(proposal, context)
-    switch (outcome) {
+    switch (outcome?.status) {
       case ProposalOutcome.REJECTED:
-        rejectedProposals.push(proposal)
+        rejectedProposals.push({ proposal, winnerChoice: outcome.winnerChoice })
```
andyesp, line 119: I think you can add this object to a variable `const proposalWithWinnerChoice = { proposal, winnerChoice: outcome.winnerChoice }` as you are using the same object in all cases.

src/services/DiscordService.ts
```diff
@@ -0,0 +1,195 @@
...
+import UpdateModel from '../entities/Updates/model'
+import { UpdateAttributes } from '../entities/Updates/types'
+import { getPublicUpdates, updateUrl } from '../entities/Updates/utils'
+
+const CHANNEL_ID = process.env.DISCORD_CHANNEL_ID
+const TOKEN = process.env.DISCORD_TOKEN
+
+const client = new Client({ intents: [GatewayIntentBits.Guilds] })
+
+const DEFAULT_AVATAR = 'https://decentraland.org/images/male.png'
```
andyesp, line 17: Don't we have a "more neutral" avatar to use as a default? To avoid using male or female avatars.

src/services/DiscordService.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (choices.length > 0) {
+      fields.push({ name: BLANK, value: BLANK }, ...choices)
+    }
+
+    const embed = new EmbedBuilder()
+      .setColor(color)
+      .setTitle(title)
+      .setURL(url)
+      .setDescription(action)
+      .setThumbnail('https://decentraland.org/images/decentraland.png')
```
andyesp, line 108: `'https://decentraland.org/images/decentraland.png'` this can be in a variable.

src/services/DiscordService.ts
```diff
@@ -0,0 +1,195 @@
...
+      action,
+      color: MessageColors.NEW_PROPOSAL,
+    })
+    this.channel.send({ embeds: [message] })
+  }
+
+  static async newUpdate(proposalId: string, proposalTitle: string, updateId: string, user: string) {
+    const publicUpdates = getPublicUpdates(await UpdateModel.find<UpdateAttributes>({ proposal_id: proposalId }))
+    const updateIdx = Number(publicUpdates.findIndex((item) => item.id === updateId))
+    const updateNumber = publicUpdates.length > 0 && updateIdx >= 0 ? publicUpdates.length - updateIdx : NaN
```
andyesp, line 157: We have this logic somewhere else (maybe the proposal updates list and in the update detail view). Do you think we can reuse it somehow?

src/services/DiscordService.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const { health, introduction, highlights, blockers, next_steps } = publicUpdates[updateIdx]
+
+    if (!health || !introduction || !highlights || !blockers || !next_steps) {
+      throw new Error('Missing update fields for Discord message')
+    }
+
+    const action = 'A new update has been created'
+    const title = `Update #${updateNumber}: ${proposalTitle}`
```
andyesp, line 166: Maybe we should handle the case this is NaN 🤔

src/services/DiscordService.ts
```diff
@@ -0,0 +1,195 @@
...
+  NEW_PROPOSAL = 0x0099ff,
+  FINISH_PROPOSAL = 0x8142f5,
+  NEW_UPDATE = 0x00ff80,
+}
+
+type EmbedMessageProps = {
+  title: string
+  proposalType?: ProposalType
+  description?: string
+  choices: Field[]
```
andyesp, line 36: I'd rename `choices` to `fields`. As they are not always choices (like in the update method)

src/services/DiscordService.ts
ncomerci: I took as reference the default avatar defined in dcl-gatsby

https://github.com/decentraland/decentraland-gatsby/blob/master/src/components/User/Avatar.tsx

ncomerci (approved):

src/services/DiscordService.ts
```diff
@@ -0,0 +1,200 @@
...
+      .setTitle(title)
+      .setURL(url)
+      .setDescription(action)
+      .setThumbnail(DCL_LOGO)
+      .addFields(...fields)
+      .setTimestamp()
+      .setFooter({ text: 'Decentraland DAO', iconURL: DCL_LOGO })
+
+    if (user) {
+      const profile = await profiles.load(user)
```
1emu, line 115: We should try/catch this, and handle the case if the profile fails to load

src/services/DiscordService.ts
```diff
@@ -0,0 +1,200 @@
...
+    name: `Option #${idx + 1}`,
+    value: capitalizeFirstLetter(choice),
+  }))
+}
+
+function shortText(text: string) {
+  return text.length > PREVIEW_MAX_LENGTH ? text.slice(0, PREVIEW_MAX_LENGTH) + '...' : text
+}
+
+export class DiscordService {
```
1emu, line 55: We should wrap all methods that call `this.channel.send`  (`finishProposal`/`newProposal`/`newUpdate`) in try/catches with proper logging (specially because we are calling this methods inside loops, and an unhandled exception could cause all following calls to fail)

src/services/DiscordService.ts
```diff
@@ -0,0 +1,200 @@
...
+  static init() {
```
1emu, line 56: Does anything guarantee us that this is called every time we create a DiscordService instance? 
Maybe it'd be better if we had this as the constructor?

src/services/DiscordService.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    client.login(TOKEN)
+  }
+
+  private static get channel() {
+    if (!CHANNEL_ID) {
+      throw new Error('Discord channel ID not set')
+    }
+
+    const channel = client.channels.cache.get(CHANNEL_ID)
```
1emu, line 69: I'm not sure about this, but I don't trust this call to always work... What if the client fails to login properly in the `init` method? 

Maybe we need to try/catch and handle

1emu (approved):
Left some comments <3. We need to handle possible errors and 🚀

src/entities/Proposal/jobs.ts
```diff
@@ -14,6 +15,16 @@ const enum ProposalOutcome {
   FINISHED = 'FINISHED',
 }
 
+type Outcome = {
+  winnerChoice: string
+  status?: ProposalOutcome
+}
+
+type ProposalWithWinnerChoice = {
```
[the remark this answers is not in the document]
andyesp: Really like this

src/pages/update.tsx
```diff
@@ -37,7 +38,7 @@ export default function UpdateDetail() {
     return <LoadingView />
   }
 
-  const index = publicUpdates && publicUpdates.length - Number(publicUpdates?.findIndex((item) => item.id === updateId))
+  const index = publicUpdates && updateId && getUpdateNumber(publicUpdates, updateId)
```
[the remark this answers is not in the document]
andyesp: Yea, not sure about returning `NaN`... Maybe return some number as a fallback, `1`?

src/services/DiscordService.ts
andyesp: This method is only called once on `server.ts`. And I think it is better that way to control exactly when we want to call it. We don't create instances of this class but use its methods. So it should always use the same client instance.

src/services/DiscordService.ts
ncomerci: the function is only calledthe function is only called when a message is sent to discord and those functions already include a trycatch, so with andy we think it is not necessary to put one here

src/entities/Proposal/jobs.ts
```diff
@@ -162,17 +157,22 @@ export async function finishProposal(context: JobContext) {
...
   }
 
   const proposals: ProposalWithWinnerChoice[] = [...finishedProposals, ...acceptedProposals, ...rejectedProposals]
   context.log(`Updating ${proposals.length} proposals in discourse... \n\n`)
-  for (const { proposal } of proposals) {
-    commentProposalUpdateInDiscourse(proposal.id)
+  for (const { id, title, winnerChoice, outcomeStatus } of proposals) {
+    commentProposalUpdateInDiscourse(id)
+    if (outcomeStatus) {
+      DiscordService.finishProposal(
```
1emu, line 170: I think we might be missing an `await`

We could wrap this with `inBackground()`, as in `commentProposalUpdateInDiscourse`

src/pages/update.tsx
```diff
@@ -37,7 +38,7 @@ export default function UpdateDetail() {
     return <LoadingView />
   }
 
-  const index = publicUpdates && publicUpdates.length - Number(publicUpdates?.findIndex((item) => item.id === updateId))
+  const index = (!!publicUpdates && !!updateId && getUpdateNumber(publicUpdates, updateId)) || NaN
```
1emu, line 41: I think this checks can be encapsulated in  `getUpdateNumber`, because it is already returning `NaN`

We could make that function to accept any input and return `NaN` if anything is not valid.

```typescript
const index = getUpdateNumber(publicUpdates, updateId))

export function getUpdateNumber(publicUpdates?: UpdateAttributes[], updateId?: string ) {
  if(!publicUpdates || !updateId) return NaN
  const updateIdx = Number(publicUpdates.findIndex((item) => item.id === updateId))
  return publicUpdates.length > 0 && updateIdx >= 0 ? publicUpdates.length - updateIdx : NaN
}
```

1emu: Also, what happens in line 50 if index is `NaN`? 
Shouldn't we handle it the same way as `updateState.error` above?

src/services/ProposalService.ts
```diff
@@ -72,6 +73,15 @@ export class ProposalService {
       coAuthors
     )
 
+    DiscordService.newProposal(
```
1emu, line 76: We could also do this `inBackground` and `await` the async method

1emu: Now that I think about it, if no one is waiting for all this `DiscordService` methods to return a value, then we could use the `inBackground(async () => { await magic }` inside each one of them, and keep calling them as we are.

WDYT? @andyesp

1emu (approved):
✨ 💪 
please check the comments on the async methods, other than that LGTM

src/services/DiscordService.ts
```diff
@@ -0,0 +1,228 @@
...
+    proposalId: string,
+    title: string,
+    type: ProposalType,
+    description: string,
+    choices: string[],
+    user: string
+  ) {
+    const action = 'A new proposal has been created'
+    const embedChoices = getChoices(choices)
+    inBackground(async () => {
```
1emu, line 152: @ncomerci we need to return a result for the inBackground successful logging.

```typescript
 .then((result) => logger.log('Completed background task', { result: JSON.stringify(result) }))
 ```

Maybe the action name, & proposal/update id

PR #622, opened 2022-12-05
fix: Address comparison bug

ncomerci:


src/entities/Votes/utils.ts
```diff
@@ -183,9 +184,11 @@ export async function getScores(
 
   const result: DetailedScores = {}
   for (const addr of formattedAddresses) {
+    const delegationScores = scores[strategies.findIndex((s) => s.name === DELEGATION_STRATEGY_NAME)] || {}
```
andyesp, line 187: You can move this out of the for loop to avoid doing this on every loop.

1emu (approved):
🕵️ ✨

andyesp (approved):

PR #666, opened 2022-12-29
Delegate VP on profile

### LINKED ISSUE #642: Add call to action to Delegate VP on Profile
When visiting your own profile:
-If you don't have a delegate: A text link under the Delegated VP to widget should open the Delegates modal.
Copy: Choose a Delegate
- If you do have a delegate: A text link on the widget should open the Delegates modal
Copy: Change your Delegate

When visiting other user's profile:
- A primary button to delegate VP to that user should appear on the profile:
Copy: Choose Delegate

1emu:
Closes #642 

When visiting your own profile:
-If you don't have a delegate: A text link under the Delegated VP to widget should open the Delegates modal.
Copy: Choose a Delegate

image

If you do have a delegate: A text link on the widget should open the Delegates modal
Copy: Change your Delegate (@ginocingolani I used guane's copies)

image

When visiting other user's profile:

A primary button to delegate VP to that user should appear on the profile:
Copy: Choose Delegate 

image

When visiting the user you delegated VP to

image

When visiting a user's profile who has not delegated but has delegators:
image

When visiting a user's profile who has delegated and has no delegators:
image

When delegating to a user:
image

image

Looking at my profile with delegation and delegators
image

When not logged in:

image

image

Looking at a profile with delegators
image


src/components/Common/Empty.tsx
```diff
@@ -17,19 +18,35 @@ interface Props {
...
-export default function Empty({ icon, title, description, className, linkText, linkHref, onLinkClick }: Props) {
+export default function Empty({
+  icon,
+  title,
+  description,
+  className,
+  linkText,
+  linkHref,
+  onLinkClick,
+  asButton,
```
andyesp, line 32: Maybe... `actionType` or `actionVariant`?
`actionVariant?: 'button' | 'link'`

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationCandidatesList.tsx
```diff
@@ -0,0 +1,84 @@
...
+      className="GovernanceContentModal VotingPowerDelegationCandidatesList"
+    >
+      <Modal.Header className="VotingPowerDelegationCandidatesList__Header">
+        {t('modal.vp_delegation.title')}
+      </Modal.Header>
+      <Modal.Description className="VotingPowerDelegationCandidatesList__Description">
+        <Markdown>
+          {(userAddress && vpDistribution
+            ? t('modal.vp_delegation.description', { vp: vpDistribution.own })
+            : t('modal.vp_delegation.description_generic')) +
```
andyesp, line 60: Are we using these two keys in other places? Or maybe we can use only one key instead of two with a space?

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationCandidatesList.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { CANDIDATE_ADDRESSES } from '../../../constants'
+import { EDIT_DELEGATE_SNAPSHOT_URL } from '../../../entities/Proposal/utils'
+import useDelegatesInfo, { Delegate } from '../../../hooks/useDelegatesInfo'
+import useVotingPowerDistribution from '../../../hooks/useVotingPowerDistribution'
+import DelegatesTable from '../../Table/DelegatesTable'
+
+import './VotingPowerDelegationCandidatesList.css'
+
+type VotingPowerDelegationCandidatesListProps = Omit<ModalProps, 'children'> & {
```
andyesp, line 18: I'd change this to just `Props`

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationModal.tsx
```diff
@@ -1,81 +1,60 @@
...
-  vision: string
-  most_important_issue: string
-}
+const VotingPowerDelegationModal = ({
+  vpDistribution,
+  openDelegationModal,
+  setOpenDelegationModal,
+  displayedCandidate,
+}: Props) => {
+  const [selectedCandidate, setSelectedCandidate] = useState<Candidate | null>(null)
```
andyesp, line 24: I'd let the parent component handle what to do with the displayedCandidate.
`VotingPowerDelegationModal` can receive an `onClose` function to call here.

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationCandidatesList.css
```diff
@@ -0,0 +1,82 @@
...
+  line-height: 36px !important;
+}
+
+.VotingPowerDelegationCandidatesList__Description {
+  margin: 0 0 32px 0 !important;
+}
+
+.VotingPowerDelegationCandidatesList > .dcl.close,
+.ui.modal > .dcl.close {
+  top: 22px; /* TODO: Can this be a property to override in all modals? */
```
ncomerci, line 15: if you want to add this to all modals, you can add it in theme.css I think

src/components/Profile/VpDelegatorsBox.tsx
```diff
@@ -15,24 +15,27 @@ import './VpDelegatorsBox.css'
...
   delegationState,
   scores,
   isLoadingScores,
-  vpDistribution,
 }: Props) {
   const t = useFormatMessage()
-  const [userAddress] = useAuthContext()
   const isLoggedUserProfile = isSameAddress(userAddress, address)
+  const { vpDistribution: loggedUserVpDistribution, isLoadingVpDistribution } = useVotingPowerInformation(
+    !isLoggedUserProfile ? userAddress : undefined
```
ncomerci, line 36: wouldn't it be better to update the hook to also accept null values and have the same behavior as undefined?

src/components/Modal/VotingPowerDelegationModal/VotingPowerDelegationCandidatesList.css
1emu: I'm not sure about this, I just moved the CSS with the comment, so I'll leave it there for now

src/components/Profile/VpDelegatorsBox.tsx
1emu: Hmm.. I don't see how that would change behavior in this case.
What I intend here is: If the profile we are looking at is not the logged user's, then load the user vp information. 
If not, the hook receives undefined and avoids calling the APIs.
 But `userAddress` could be a valid value, and still we would want to send `undefined` to the hook.

andyesp (approved):

src/components/Profile/VpDelegatorsBox.tsx
ncomerci: aaah okay, I wasn't contemplating the last case you described. Okay then!

PR #691, opened 2023-01-17
feat: grant submit changes

andyesp:


src/components/GrantRequest/GrantRequestFundingSection.tsx
```diff
@@ -142,139 +146,85 @@ export default function GrantRequestFundingSection({ onValidation, isFormDisable
...
+              value={state.value.funding}
+              onChange={({ currentTarget }) =>
+                editor.set({
+                  funding: currentTarget.value,
+                  projectDuration: updateProjectDuration(currentTarget.value, state.value.projectDuration),
+                })
+              }
+              onBlur={() => editor.set({ funding: state.value.funding })}
+              error={state.error.funding || ''}
+              // message={t(state.error.funding)}
```
1emu, line 199: we can remove this :)

src/components/GrantRequest/ProjectDurationInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }, [onChange, optionMinLimit, value])
+
+  return (
+    <div>
+      <Label>{t('page.submit_grant.funding_section.project_duration_title')}</Label>
+      <div>
+        <div className="ProjectDurationInput__InputContainer">
+          <div className="ProjectDurationInput__Input">
+            <button className="ProjectDurationInput__Button" onClick={handleRemoveClick}>
+              <Remove />
```
1emu, line 45: WDYT about increase/decrease instead of add/remove?

1emu (approved):
👏👏👏

src/components/GrantRequest/ProjectDurationInput.tsx
andyesp: I think those are the common names in the Icons Realm. Other possibility could be using Minus / Plus

andyesp: I like better Increase/Decrease for this case. But I don't know if it's better to keep them more generic.

PR #692, opened 2023-01-18
New grants page

ncomerci:


src/components/Common/ProgressBar.tsx
```diff
@@ -0,0 +1,21 @@
...
+  height: string
+  background: string
+}
+
+function ProgressBar({ percentage, height, background }: Props) {
+  const width = `${percentage < 0 ? 0 : percentage > 100 ? 100 : percentage}%`
+
+  return (
+    <div className="ProgressBar__Container" style={{ height }}>
+      <div className="ProgressStatus" style={{ width, background }}></div>
```
andyesp, line 16: This should be `ProgressBar__Status`

src/components/Grants/BannerItem.css
```diff
@@ -5,6 +5,11 @@
   justify-content: space-between;
 }
 
+.GrantsBannerItem--noUrl {
```
andyesp, line 8: Maybe doing the opposite and changing this to `.GrantsBannerItem--clickable`?

src/components/Grants/BannerItem.tsx
```diff
@@ -1,22 +1,39 @@
...
     <>
-      <div className="GrantsBannerItem">
-        <div>
-          <h3 className="GrantsBannerItem__Title">{title}</h3>
-          <p className="GrantsBannerItem__Description">{description}</p>
+      <a
+        href={url || ''}
+        target="_blank"
+        className={TokenList.join([!url && 'GrantsBannerItem--noUrl'])}
+        rel="noreferrer"
```
andyesp, line 23: I think this should be `rel="noopener noreferrer"`

src/components/Grants/RequestBanner.css
```diff
@@ -0,0 +1,39 @@
...
+  font-style: normal;
+  font-weight: 400;
+  font-size: 15px;
+  line-height: 22px;
+  color: var(--black-600);
+  margin-bottom: 16px;
+}
+
+.ui.fluid.button.RequestBanner__Button {
+  background: #ffffff;
```
andyesp, line 29: We have a white value to use here.

src/components/Grants/RequestBanner.css
```diff
@@ -0,0 +1,39 @@
+.RequestBanner {
+  background: linear-gradient(89.29deg, #f6f5f8 0%, rgba(246, 245, 248, 0) 100%);
```
andyesp, line 2: Don't we have these colors in the theme?

src/components/Grants/Current/CurrentGrantsBanner.tsx
```diff
@@ -1,59 +1,30 @@
...
-    totalToBeVested,
-  }
-}
-
-const CurrentGrantsBanner = ({ grants }: { grants: GrantWithUpdateAttributes[] }) => {
+const CurrentGrantsBanner = () => {
   const t = useFormatMessage()
-  const bannerStats = useMemo(() => getBannerStats(grants), [grants])
-  const bannerItems = useMemo(
+  const bannerItems = useMemo<BannerItem[]>(
```
andyesp, line 9: If we don't have stats now we can remove this memo and this array and put all those texts directly on the render.

src/components/Grants/Current/BudgetBannerItem.tsx
```diff
@@ -0,0 +1,19 @@
...
+import './BudgetBannerItem.css'
+
+interface Props {
+  value: string
+  label: string
+}
+
+function BudgetBannerItem({ value, label }: Props) {
+  return (
+    <div className="BudgetBannerItem">
```
andyesp, line 12: This class is not necessary :)

src/components/Grants/Current/BudgetBanner.css
```diff
@@ -0,0 +1,39 @@
+.BudgetBanner {
+  box-sizing: border-box;
+  background: linear-gradient(89.29deg, #f6f5f8 0%, rgba(246, 245, 248, 0) 100%);
```
andyesp, line 3: Are we reusing this gradient? Maybe we can move it to the theme

andyesp (changes requested):

src/components/Grants/Current/BudgetBanner.css
1emu: Good idea, it is the same as the [RequestBanner.css](https://github.com/decentraland/governance/pull/692/files#diff-70ed416cd8c009a1f0f435613fbfc8a6c7f5960805946b6e60e55529c4dd3900)

src/components/Search/StatusFilter.tsx
```diff
@@ -2,39 +2,48 @@ import React, { useMemo } from 'react'
...
+  statusType,
+}: FilterProps & { statusType: StatusType }) {
   const t = useFormatMessage()
   const location = useLocation()
   const params = useMemo(() => new URLSearchParams(location.search), [location.search])
-  const status = toProposalStatus(params.get('status')) ?? undefined
+  const status = params.get('status')
 
-  function handleStatusFilter(status: ProposalStatus | null) {
+  function handleStatusFilter(status: string | null) {
```
1emu, line 27: This is really similar to

```typescript
function handleTypeFilter(type: string | null, params: URLSearchParams) {
  const newParams = new URLSearchParams(params)
  type ? newParams.set('type', type) : newParams.delete('type')
  newParams.delete('page')
  const stringParams = newParams.toString()
  return `${location.pathname}${stringParams === '' ? '' : '?' + stringParams}`
}
```

I think the only difference is the "filter key". 
Could you extract it to a common fn? 🙏 
I think we could have a file for common functions to filters.

You are also using `Object.values` below, and there's this function defined in the other filter, that could be reused:

```typescript
function getValues<T extends object>(e: T): Array<T[keyof T]> {
  return Object.values(e)
}
```

src/pages/grants.tsx
```diff
@@ -1,23 +1,48 @@
...
 import Navigation, { NavigationTab } from '../components/Layout/Navigation'
+import CategoryFilter from '../components/Search/CategoryFilter'
+import StatusFilter from '../components/Search/StatusFilter'
+import { GrantStatus, NewGrantCategory, OldGrantCategory } from '../entities/Grant/types'
+import { toGrantStatus, toProposalGrantCategory } from '../entities/Grant/utils'
+import { GrantWithUpdateAttributes } from '../entities/Proposal/types'
 import useGrants from '../hooks/useGrants'
 import { isUnderMaintenance } from '../modules/maintenance'
 
+function filterDisplayableGrants(grants: GrantWithUpdateAttributes[], type: string | null, status: string | null) {
```
1emu, line 27: I think we could write this like this to avoid going through all grants twice:

```typescript
function filterDisplayableGrants(grants: GrantWithUpdateAttributes[], type: string | null, status: string | null) {
  return status || type
    ? grants.filter(
        (grant) =>
          (type ? toSnakeCase(grant.configuration.category) === type : true) &&
          (status ? toSnakeCase(grant.status) === status : true)
      )
    : grants
}
```

1emu (approved):
Nice one! I'm ok to merge after addressing comments :)

src/components/Grants/BannerItem.tsx
ncomerci: image

src/components/Grants/RequestBanner.css
ncomerci: nope :(

src/components/Grants/BannerItem.css
ncomerci: The default behavior is clickable, I need it to be non-clickable 😅

src/clients/utils.ts
```diff
@@ -36,3 +36,11 @@ export function trimLastForwardSlash(url: string) {
 export function capitalizeFirstLetter(string: string) {
   return string.length > 0 ? `${string[0].toUpperCase()}${string.slice(1)}` : ''
 }
+
+export function handleUrlFilters<T>(filterKey: string, value: T | null, params: URLSearchParams) {
```
1emu, line 40: Love this! Maybe switch `value` and `params` so value can be optional and we don't need to be calling the function with `null`?

src/entities/Proposal/types.ts
```diff
@@ -484,15 +468,70 @@ export type GrantProposalConfiguration = GrantRequestGeneralInfo & {
...
-export const VALID_CATEGORIES = [
-  ProposalGrantCategory.Accelerator,
-  ProposalGrantCategory.CoreUnit,
-  ProposalGrantCategory.Documentation,
-  ProposalGrantCategory.InWorldContent,
-  ProposalGrantCategory.Platform,
-  ProposalGrantCategory.SocialMediaContent,
-  ProposalGrantCategory.Sponsorship,
-]
+export const newProposalGrantScheme = {
```
1emu, line 471: This we don't need anymore I think

1emu (approved):
Nice Nicoo. Left some minor comments. Ok to merge after that :)

1emu (approved):

PR #698, opened 2023-01-20
feat: new grants

andyesp:


ncomerci (commented):
If the form submission fails, no error message is displayed

.env.example
```diff
@@ -45,23 +44,7 @@ GATSBY_VOTING_POWER_TO_PASS_DRAFT=0
...
-# Override VP for Grant Tiers, defaults to GATSBY_VOTING_POWER_TO_PASS_GRANT
-GATSBY_VOTING_POWER_TO_PASS_GRANT_TIER4=600
-GATSBY_VOTING_POWER_TO_PASS_GRANT_TIER5=700
-GATSBY_VOTING_POWER_TO_PASS_GRANT_TIER6=800
-
-# Override Duration for Grant Tiers, defaults to GATSBY_SNAPSHOT_DURATION
-GATSBY_DURATION_GRANT_TIER4=1200
-GATSBY_DURATION_GRANT_TIER5=1200
-GATSBY_DURATION_GRANT_TIER6=1200
+GATSBY_GRANT_SIZE_MINIMUM=1
```
andyesp, line 47: We changed the minimum to 100, right?

1emu: We are not using this constant anymore, we replaced it by:

image

We could add them to the configs json but it's not like we need different configurations per environment. So we could just remove it and leave those there, WDYT?

1emu: https://github.com/decentraland/governance/pull/740

1emu (approved):
LFG!

PR #702, opened 2023-01-23
Submit improvements

1emu:
- Show loader on submit
- Validate general info section after adding a coauthor
- Switch funding order (experimental)

### DIFF
--- a/src/pages/submit/grant.tsx
+++ b/src/pages/submit/grant.tsx
@@ -69,9 +69,13 @@ export default function SubmitGrant() {
   const isCategorySelected = grantRequest.category !== null
   const preventNavigation = useRef(false)
 
-  const submit = () => {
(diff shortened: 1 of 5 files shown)


src/pages/submit/grant.tsx
```diff
@@ -90,15 +92,8 @@ export default function SubmitGrant() {
...
         })
     }
-    setIsFormDisabled(false)
   }
 
-  useEffect(() => {
-    preventNavigation.current = userModifiedForm(grantRequest, initialState)
-  }, [grantRequest])
-
-  usePreventNavigation(!!preventNavigation)
```
andyesp, line 100: Don't we need this hook?

1emu: 🙈

PR #703, opened 2023-01-23
feat: vesting start date radio group

andyesp:


1emu (changes requested):
I think this needs to be added to `GrantProposalConfiguration` so we save the data to the db

src/entities/Proposal/types.ts
```diff
@@ -465,6 +465,7 @@ export type GrantProposalConfiguration = GrantRequestGeneralInfo & {
   projectDuration?: number // Old grants may not have this field
   tier: GrantTierType
   choices: string[]
+  vestingStartDate: VestingStartDate
```
1emu, line 468: this needs to be optional since old grants won't have this field

1emu (changes requested):

1emu (approved):

PR #708, opened 2023-01-31
Budget banner changes

ncomerci:


1emu (approved):

src/hooks/useBudgetByCategory.ts
```diff
@@ -1,8 +1,56 @@
...
+
 import { PROPOSAL_GRANT_CATEGORY_ALL } from '../entities/Proposal/types'
 
+import { Governance } from './../clients/Governance'
 import { ProposalGrantCategory } from './../entities/Grant/types'
 
+type Budget = {
+  percentage: number
+  currentAmount: number
+  totalBudget: number
```
1emu, line 14: I'd rather we call this properties `allocated` and `total`.

Also, I think the `percentage` name could be confusing depending on the context. How about `allocatedPercentage` ?

1emu (approved):

PR #709, opened 2023-01-31
Feat/fetch budgets

1emu:
image

image

### DIFF
--- a/src/clients/Governance.ts
+++ b/src/clients/Governance.ts
@@ -21,10 +21,13 @@ import {
   ProposalStatus,
   ProposalType,
 } from '../entities/Proposal/types'
+import { QuarterBudgetAttributes } from '../entities/QuarterBudget/types'
 import { SubscriptionAttributes } from '../entities/Subscription/types'
 import { ProjectHealth, UpdateAttributes } from '../entities/Updates/types'
 import { Vote, VotedProposal } from '../entities/Votes/types'
 import { GrantRequest } from '../pages/submit/grant'
+
+import { TransparencyBudget } from './DclData'
 
 type NewProposalMap = {
   [`/proposals/poll`]: NewProposalPoll
@@ -360,6 +363,19 @@ export class Governance extends API {
     const response = await this.fetch<ApiResponse<{ total: number; available: number; allocated: number }>>(
       `/budget/${snakeCase(category)}`,
       this.options().method('GET')
+    )
+    return response.data
+  }
+
+  async getTransparencyBudgets() {
+    const response = await this.fetch<ApiResponse<TransparencyBudget[]>>(`/budget/fetch`, this.options().method('GET'))
+    return response.data
+  }
+
+  async updateGovernanceBudgets() {
+    const response = await this.fetch<ApiResponse<QuarterBudgetAttributes[]>>(
+      `/budget/update`,
+      this.options().method('GET').authorization({ sign: true })
     )
     return response.data
   }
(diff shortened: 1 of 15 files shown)


src/clients/Governance.ts
```diff
@@ -363,4 +366,17 @@ export class Governance extends API {
...
+
+  async getTransparencyBudgets() {
+    const response = await this.fetch<ApiResponse<TransparencyBudget[]>>(`/budget/fetch`, this.options().method('GET'))
+    return response.data
+  }
+
+  async updateGovernanceBudgets() {
+    const response = await this.fetch<ApiResponse<QuarterBudgetAttributes[]>>(
+      `/budget/update`,
+      this.options().method('GET').authorization({ sign: true })
```
ncomerci, line 378: Since this endpoint modifies the database, I would use a POST or PATCH instead of a GET

src/entities/Budget/routes.ts
```diff
@@ -45,31 +50,13 @@ export const BUDGET = {
...
-      available: 0,
-      allocated: 0,
-    },
   },
 }
 
 export default routes((route) => {
+  const withAuth = auth()
+  route.get('/budget/fetch/', handleAPI(fetchBudgets))
+  route.get('/budget/update/', withAuth, handleAPI(updateBudgets))
```
ncomerci, line 59: Related to the above

src/entities/QuarterBudget/model.ts
```diff
@@ -0,0 +1,67 @@
...
+  static async createNewBudgets(transparencyBudgets: TransparencyBudget[]) {
+    const existingBudgets = await this.find<QuarterBudgetAttributes>()
+    const newBudgets: QuarterBudgetAttributes[] = []
+    if (transparencyBudgets.length > existingBudgets.length) {
+      const sortedBudgets = transparencyBudgets.sort((a, b) => Date.parse(b.start_date) - Date.parse(a.start_date))
+      const now = new Date()
+      for (const transparencyBudget of sortedBudgets) {
+        const startAt = new Date(transparencyBudget.start_date)
+        if (!existingBudgets.some((existingBudget) => existingBudget.start_at === startAt)) {
+          const finishAt = Time(transparencyBudget.start_date).add(4, 'months').toDate() //TODO a chequear
```
ncomerci, line 26: Checked?

src/entities/QuarterCategoryBudget/utils.ts
```diff
@@ -0,0 +1,38 @@
+import { snakeCase } from 'lodash'
+
+import { TransparencyBudget } from '../../clients/DclData'
+import { NewGrantCategory } from '../Grant/types'
+import { QuarterBudgetAttributes } from '../QuarterBudget/types'
+
+export function toNewGrantCategory(category: string): NewGrantCategory {
```
ncomerci, line 7: May I suggest this to avoid changing the code if a new category is added?

```typescript
function toNewGrantCategory(category: string) {
  const categories = Object.values(NewGrantCategory)
  const idx = categories.map(snakeCase).findIndex((value) => value === snakeCase(category))
  
  if(idx >= 0) {
    return categories[idx]
  }

  throw Error(`Attempted to parse an invalid NewGrantCategory ${category}`)
}
```

ncomerci (approved):

src/entities/QuarterBudget/model.ts
1emu: I'm adding this to the tests in the next PR :)

PR #719, opened 2023-02-07
feat: Added coauthors list option to proposals API endpoint

ncomerci:
When the query param `coauthor=true` is added to the `/proposals` endpoint, the list of approved co-authors (if any) is displayed.

This is a required modification in order to display the list of coauthors in the transparency proposals sheet.

image

### DIFF
--- a/src/entities/Proposal/model.ts
+++ b/src/entities/Proposal/model.ts
@@ -240,7 +240,9 @@ export default class ProposalModel extends Model<ProposalAttributes> {
     return (!!result && result[0] && Number(result[0].total)) || 0
   }
 
-  static async getProposalList(filter: Partial<FilterProposalList & FilterPagination> = {}) {
+  static async getProposalList(
+    filter: Partial<FilterProposalList & FilterPagination> = {}
+  ): Promise<(ProposalAttributes & { coauthors?: string[] | null })[]> {
     if (filter.user && !isEthereumAddress(filter.user)) {
       return []
     }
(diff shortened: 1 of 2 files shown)


src/entities/Proposal/model.ts
```diff
@@ -291,7 +307,7 @@ export default class ProposalModel extends Model<ProposalAttributes> {
       SQL`AND p."finish_at" > NOW() AND p."finish_at" < ${timeFrame}`
     )}
     ${conditional(!!filter.search, SQL`AND "rank" > 0`)}
-    ORDER BY ${conditional(!!filter.coauthor, SQL`CASE c.status WHEN 'PENDING' THEN 1 END,`)} 
+    ORDER BY ${conditional(!!filter.coauthor && !!filter.user, SQL`CASE c.status WHEN 'PENDING' THEN 1 END,`)}
```
1emu, line 310: Do you want this order by only if there is a user?

src/entities/Proposal/types.ts
```diff
@@ -63,6 +63,7 @@ export type ProposalAttributes<C extends Record<string, unknown> = any> = {
   created_at: Date
   updated_at: Date
   textsearch: SQLStatement | string | null | undefined
+  coauthors?: string[] | null
```
1emu, line 66: I'm not sure if we should add this field here if it's not going to be a column in the db.

Perhaps you could just change the return type of the methods in the model to

`ProposalAttributes & {coauthors?: string[] | null}`

1emu (approved):

PR #723, opened 2023-02-08
feat: add out of budget to proposal status

1emu:
Not sure if we should be able to link grant proposals with OOB status, but if the idea is to be able to resubmit and link to them, I think we should.

Moved frontend behavior depending on `ProposalStatus` to tested functions in `utils.ts`, so it's easier to refactor them later.

### DIFF
--- a/src/entities/Proposal/templates/messages.ts
+++ b/src/entities/Proposal/templates/messages.ts
@@ -3,8 +3,7 @@ import { calculateResult } from '../../Votes/utils'
 import { ProposalAttributes, ProposalStatus } from '../types'
 
 function getProposalStatusDisplayName(proposalStatus: ProposalStatus) {
-  if (proposalStatus === ProposalStatus.OutOfBudget) return 'OUT OF BUDGET'
-  return proposalStatus.toUpperCase()
+  return proposalStatus.split('_').join(' ').toUpperCase()
 }
 
 export function getUpdateMessage(proposal: ProposalAttributes, votes: Record<string, Vote>) {
--- a/src/entities/Proposal/templates/messages.ts
+++ b/src/entities/Proposal/templates/messages.ts
@@ -2,10 +2,15 @@ import { Vote } from '../../Votes/types'
 import { calculateResult } from '../../Votes/utils'
 import { ProposalAttributes, ProposalStatus } from '../types'
 
+function getProposalStatusDisplayName(proposalStatus: ProposalStatus) {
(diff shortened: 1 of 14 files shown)


src/entities/Proposal/utils.ts
```diff
@@ -163,3 +149,39 @@ export function isProposalInCliffPeriod(grant: TransparencyGrant) {
...
+    default:
+      return false
+  }
+}
+
+export function toProposalStatus(value: string | null | undefined, orElse: () => any): ProposalStatus | any {
+  return isProposalStatus(value) ? (value as ProposalStatus) : orElse()
+}
+
+export function proposalCanBeDeleted(proposalStatus?: ProposalStatus) {
```
andyesp, line 173: `isProposalDeletable`?

src/entities/Proposal/utils.ts
```diff
@@ -163,3 +149,39 @@ export function isProposalInCliffPeriod(grant: TransparencyGrant) {
...
+  return proposalStatus === ProposalStatus.Pending || proposalStatus === ProposalStatus.Active
+}
+
+export function proposalCanBeEnacted(proposalStatus?: ProposalStatus) {
```
andyesp, line 177: `isProposalEnactable`?

src/entities/Proposal/templates/messages.ts
```diff
@@ -2,10 +2,15 @@ import { Vote } from '../../Votes/types'
 import { calculateResult } from '../../Votes/utils'
 import { ProposalAttributes, ProposalStatus } from '../types'
 
+function getProposalStatusDisplayName(proposalStatus: ProposalStatus) {
+  if (proposalStatus === ProposalStatus.OutOfBudget) return 'OUT OF BUDGET'
```
andyesp, line 6: You can do something like `ProposalStatus.OutOfBudget.split('_').join(' ').toUpperCase()`

src/intl/en.json
```diff
@@ -286,7 +286,8 @@
     "active": "Active outcomes",
     "rejected": "Rejected outcomes",
     "passed": "Passed outcomes",
-    "enacted": "Enacted outcomes"
+    "enacted": "Enacted outcomes",
+    "out_of_budget": "OOB outcomes"
```
andyesp, line 290: Is this ok with @ginocingolani? Not sure about the acronym here.

1emu: I believe Guane is going to use the acronym in the pills. But we can change it later :)

PR #728, opened 2023-02-14
feat: FinishProposal Job Update

1emu:


src/clients/Discourse.ts
```diff
@@ -295,10 +293,11 @@ export class Discourse extends API {
 
   constructor(baseUrl: string) {
     super(baseUrl)
-    this.auth = this.getCredentials()
+    this.auth = Discourse.getCredentials()
   }
 
-  private getCredentials() {
+  private static getCredentials() {
+    const DISCOURSE_API_KEY = requiredEnv('DISCOURSE_API_KEY')
```
andyesp, line 300: Isn't this already throwing an error if env is not present? If this is the case then condition in line 301 is not necessary anymore

src/clients/Governance.ts
```diff
@@ -385,4 +385,12 @@ export class Governance extends API {
     )
     return response.data
   }
+
+  async reportErrorToServer(message: string) {
+    const response = await this.fetch<ApiResponse<string>>(
+      `/debug/report_error`,
```
andyesp, line 391: I think we are using kebab case for endpoints routes (`report-error`)

src/config/env/dev.json
```diff
@@ -36,5 +36,6 @@
   "GATSBY_VOTING_POWER_TO_PASS_GOVERNANCE": "500",
   "GATSBY_VOTING_POWER_TO_PASS_POI": "99",
   "GATSBY_VOTING_POWER_TO_PASS_POLL": "500",
-  "GATSBY_GRANT_SIZE_MINIMUM": "1"
+  "GATSBY_GRANT_SIZE_MINIMUM": "1",
+  "DISCORD_SERVICE_ENABLED": "false"
```
andyesp, line 40: This variable is not necessary here as the discord service is only used server side 🤔

src/config/env/heroku.json
```diff
@@ -35,5 +35,6 @@
   "GATSBY_VOTING_POWER_TO_PASS_GOVERNANCE": "500",
   "GATSBY_VOTING_POWER_TO_PASS_POI": "10000",
   "GATSBY_VOTING_POWER_TO_PASS_POLL": "500",
-  "GATSBY_GRANT_SIZE_MINIMUM": "1"
+  "GATSBY_GRANT_SIZE_MINIMUM": "1",
+  "DISCORD_SERVICE_ENABLED": "false"
```
andyesp, line 39: This variable is not necessary here as the discord service is only used server side 🤔

src/config/env/local.json
```diff
@@ -33,5 +33,7 @@
   "GATSBY_SUBMISSION_THRESHOLD_DRAFT": "100",
   "GATSBY_SUBMISSION_THRESHOLD_GOVERNANCE": "100",
   "GATSBY_SEGMENT_KEY": "LOCAL_ENVIRONMENT_VARIABLE",
-  "GATSBY_LOCAL_ENV_VAR": "only for testing local"
+  "GATSBY_LOCAL_ENV_VAR": "only for testing local",
+  "DISCORD_SERVICE_ENABLED": "false",
```
andyesp, line 37: This variable is not necessary here as the discord service is only used server side 🤔

src/config/env/prd.json
```diff
@@ -36,5 +36,6 @@
   "GATSBY_VOTING_POWER_TO_PASS_GOVERNANCE": "6000000",
   "GATSBY_VOTING_POWER_TO_PASS_POI": "500000",
   "GATSBY_VOTING_POWER_TO_PASS_POLL": "500000",
-  "GATSBY_GRANT_SIZE_MINIMUM": "1"
+  "GATSBY_GRANT_SIZE_MINIMUM": "1",
+  "DISCORD_SERVICE_ENABLED": "true"
```
andyesp, line 40: This variable is not necessary here as the discord service is only used server side 🤔

src/entities/Debug/routes.ts
```diff
@@ -1,11 +1,21 @@
...
 import { DEBUG_ADDRESSES } from './isDebugAddress'
 
 export default routes((router) => {
-  return router.get(
+  const withAuth = auth()
+  router.get(
     '/debug',
     handleAPI(async () => DEBUG_ADDRESSES)
   )
+  router.post('/debug/report_error', withAuth, handleAPI(reportClientError))
```
andyesp, line 16: Same here, we are naming endpoints in kebab case.

src/services/DiscourseService.ts
```diff
@@ -49,7 +49,7 @@ export class DiscourseService {
     }
 
     return {
-      category: DISCOURSE_CATEGORY ? Number(DISCOURSE_CATEGORY) : undefined,
+      category: DiscourseService.DISCOURSE_CATEGORY ? Number(DiscourseService.DISCOURSE_CATEGORY) : undefined,
```
andyesp, line 52: If `DISCOURSE_CATEGORY` is a required env for the server to run then this condition is not necessary.

```
return {
  category: Number(DISCOURSE_CATEGORY)
  ...
}
```

But maybe typescript throws an error here, not sure what requiredEnv returns.

src/entities/Proposal/jobs.ts
```diff
@@ -1,168 +1,169 @@
...
-import { SnapshotGraphql } from '../../clients/SnapshotGraphql'
+import { BudgetService } from '../../services/BudgetService'
 import { DiscordService } from '../../services/DiscordService'
+import { ErrorService } from '../../services/ErrorService'
+import { CurrentBudget } from '../Budget/types'
+import { BUDGETING_START_DATE } from '../Grant/constants'
 import UpdateModel from '../Updates/model'
-import { Scores } from '../Votes/utils'
 
+import { Outcome, ProposalOutcome, calculateOutcome } from './calculateOutcome'
```
andyesp, line 11: WDYT if we rename this file to `outcome`?

src/entities/Proposal/jobs.test.ts
```diff
@@ -0,0 +1,344 @@
...
+import ProposalModel from './model'
+import * as routes from './routes'
+import {
+  ACCEPTED_OUTCOME,
+  FINISHED_OUTCOME,
+  JOB_CONTEXT_MOCK,
+  REJECTED_OUTCOME,
+  createTestProposal,
+  getTestBudgetWithAvailableSize,
+} from './test_helpers'
```
andyesp, line 23: As a convention we are not using snake case for files. Use camelCase instead :)

src/entities/QuarterBudget/model.ts
```diff
@@ -104,12 +104,38 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
     `
 
     const result = await this.query(query)
+    if (!result || result.length === 0) {
+      return null
+    }
+    return this.parseCurrentBudget(result)
+  }
+
+  static async getBudget(dateWithinBudget: Date): Promise<CurrentBudget | null> {
```
andyesp, line 113: `getBudgetFromDate`?

src/services/BudgetService.ts
```diff
@@ -130,4 +150,40 @@ export class BudgetService {
...
+    minDate?: Date
+    maxDate?: Date
+  } {
+    const sorted = proposals
+      .filter((p) => p.start_at && p.start_at >= BUDGETING_START_DATE)
+      .sort((p1, p2) => p1.start_at.getTime() - p2.start_at.getTime())
+    return { minDate: sorted[0]?.start_at, maxDate: sorted[sorted.length - 1]?.start_at }
+  }
+
+  static async getBudgets(proposals: Pick<ProposalAttributes, 'start_at'>[]): Promise<CurrentBudget[]> {
```
andyesp, line 164: This will only return two budgets right? What if we return an object like `{ current: newestBudget, previous: previousBudget }`?

(Renaming oldestBudget to previousBudget)

src/services/BudgetService.ts
```diff
@@ -107,14 +116,25 @@ export class BudgetService {
     return await QuarterBudgetModel.createNewBudgets(transparencyBudgets)
   }
 
-  static getCurrentBudget(): Promise<CurrentBudget> {
-    return QuarterBudgetModel.getCurrentBudget()
+  static async getCurrentBudget(): Promise<CurrentBudget> {
+    const currentBudget = await QuarterBudgetModel.getCurrentBudget()
+    if (currentBudget !== null) {
+      return currentBudget
+    } else {
```
andyesp, line 123: You can remove the else condition as you are doing a return in the if condition.

src/services/BudgetService.ts
```diff
@@ -107,14 +116,25 @@ export class BudgetService {
...
   }
 
   static async getCategoryBudget(category: NewGrantCategory): Promise<CurrentCategoryBudget> {
     const categoryBudget = await QuarterBudgetModel.getCategoryBudgetForCurrentQuarter(category)
-    const { total, allocated } = categoryBudget
-    return { total, allocated, available: total - allocated }
+    if (categoryBudget !== null) {
+      const { total, allocated } = categoryBudget
+      return { total, allocated, available: total - allocated }
+    } else {
```
andyesp, line 134: You can remove the `else` here too.

src/services/BudgetService.ts
andyesp: This name makes me think the function returns ALL budgets. Maybe `getLatestBudgets`?

src/services/BudgetService.ts
```diff
@@ -130,4 +150,40 @@ export class BudgetService {
...
+    const oldestBudget = await QuarterBudgetModel.getBudget(minDate)
+    if (oldestBudget === null) {
+      ErrorService.report(`Could not find budget for ${minDate}`)
+      return budgetsForProposals
+    }
+    budgetsForProposals.push(oldestBudget)
+    if (maxDate && minDate !== maxDate) {
+      const newestBudget = await QuarterBudgetModel.getBudget(maxDate)
+      if (newestBudget !== null) {
+        const index = budgetsForProposals.findIndex((budget) => budget.id === newestBudget.id)
```
andyesp, line 177: Is this for checking if newestBudget is the same is the oldestBudget?
What if you do something like...
```
const newestBudget = await QuarterBudgetModel.getBudget(maxDate)
if(newestBudget !== null && newestBudget.id !== oldestBudget.id) {
  budgetsForProposals.push(newestBudget)
}
```

src/services/BudgetService.ts
```diff
@@ -130,4 +150,40 @@ export class BudgetService {
...
+        if (index === -1) budgetsForProposals.push(newestBudget)
+      }
+    }
+    return budgetsForProposals
+  }
+
+  static async updateBudgets(budgetUpdates: CurrentBudget[]) {
```
andyesp, line 184: I'd rename param to `budgets`.

src/services/BudgetService.ts
```diff
@@ -130,4 +150,40 @@ export class BudgetService {
       )
     }
   }
+
+  static getProposalsBudgetingMinAndMaxDates(proposals: Pick<ProposalAttributes, 'start_at'>[]): {
```
andyesp, line 154: I think we can drop `budgeting` from the name function 🤔

src/entities/Proposal/calculateOutcome.ts
```diff
@@ -0,0 +1,95 @@
+import { Env } from '@dcl/ui-env'
+import JobContext from 'decentraland-gatsby/dist/entities/Job/context'
+
+import { SnapshotGraphql } from '../../clients/SnapshotGraphql'
+import { config } from '../../config'
+import { Scores } from '../Votes/utils'
+
+import { INVALID_PROPOSAL_POLL_OPTIONS, ProposalAttributes } from './types'
+
+function sameOptions(options: string[], expected: string[]) {
```
andyesp, line 10: What if we use `isEqual` from lodash?

src/entities/Proposal/jobs.ts
```diff
@@ -1,168 +1,169 @@
...
-        (isForAgainst && results['for'] > results['against'])
-      ) {
-        outcome.outcomeStatus = ProposalOutcome.ACCEPTED
-      } else {
-        outcome.outcomeStatus = ProposalOutcome.REJECTED
-      }
-    } else {
-      outcome.outcomeStatus = ProposalOutcome.FINISHED
-    }
+function getCategoryBudgetFor(proposal: ProposalWithOutcome, proposalBudget: CurrentBudget) {
```
andyesp, line 76: For...? 😛

src/entities/Proposal/jobs.ts
```diff
@@ -1,168 +1,169 @@
...
-  const pendingProposals = await ProposalModel.getFinishedProposals()
-  if (pendingProposals.length === 0) {
-    context.log(`No finished proposals...`)
-    return
-  }
+function updateCategoryBudget(proposal: ProposalWithOutcome, proposalBudget: CurrentBudget) {
+  const categoryBudget = getCategoryBudgetFor(proposal, proposalBudget)
+  const size = asNumber(proposal.configuration.size)
+  categoryBudget.allocated += size
+  categoryBudget.available -= size
```
andyesp, line 91: I think these two are not necessary as you are not using categoryBudget anymore 🤔

src/services/DiscourseService.ts
1emu: 🤔 
It should always return a string, but if it's an empty string, it would be turned into `0`
I don't know if requiredEnv throws if you have an empty string, but I'd rather not change this right now :)

src/services/BudgetService.ts
1emu: Yeah, the thing is that it won't just return you the min and max dates for every proposal, it filters those that started after the budgeting system started. I thought the name should give you a hint

src/entities/Proposal/calculateOutcome.ts
1emu: Could be, but I'd rather not change this now, I just moved it :)

PR #753, opened 2023-02-23
refactor: add multiple choice field component

andyesp:


1emu (approved):
Remove `console.log('s', state.value.eventCategory)` and mergee

PR #777, opened 2023-03-02
feat: Grant forms' own url

ncomerci:
image

### DIFF
--- a/src/pages/submit/grant.tsx
+++ b/src/pages/submit/grant.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import Helmet from 'react-helmet'
 
 import Head from 'decentraland-gatsby/dist/components/Head/Head'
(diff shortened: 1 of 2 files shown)


src/pages/submit/grant.tsx
```diff
@@ -79,9 +80,30 @@ function parseStringsAsNumbers(grantRequest: GrantRequest) {
...
+  }
+}
+
 export default function SubmitGrant() {
   const t = useFormatMessage()
   const [account, accountState] = useAuthContext()
+  useEffect(() => {
+    const params = new URLSearchParams(location.search)
+    let category: NewGrantCategory | null = null
+    try {
```
andyesp, line 97: Why don't...?
```
    try {
      const category = toNewGrantCategory(params.get('category'))
      patchGrantRequest({ category })
    } catch (error) {
      console.error(error)
    }
```

src/pages/submit/grant.tsx
```diff
@@ -79,9 +80,30 @@ function parseStringsAsNumbers(grantRequest: GrantRequest) {
...
+      category = toNewGrantCategory(params.get('category'))
+    } catch (error) {
+      console.error(error)
+    } finally {
+      patchGrantRequest({ category })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
andyesp, line 104: I'd remove this and add the dependencies. Is params triggering a re-render?

src/pages/submit/grant.tsx
ncomerci: If `toNewGrantCategory` throws an error, the patch would not be executed and it is necessary that it is done regardless of whether it fails or not, that is why it is in finally

src/pages/submit/grant.tsx
ncomerci: image

It suggests adding the patch function to the dependencies, which is not necessary. That's why I omitted it

1emu: Don't you need to add `location.search` to the deps?

andyesp (approved):

PR #817, opened 2023-03-16
feat: update and add budget breakdown section texts

andyesp:


src/components/GrantRequest/GrantRequestSection.css
```diff
@@ -84,11 +84,13 @@
 .GrantRequestSection__Field--row {
```
1emu, line 84: If 'row' is not a modifier to the CSS class we should call it `__FieldRow`, right?

1emu (approved):

PR #842, opened 2023-03-21
feat: hide voting results and update styles

### LINKED ISSUE #806: Add copy/heading to vote component + "Hide" partial results
- Add copy before displaying vote options
- Copy to be displayed, at first, only for Grant Requests
- Acceptance threshold is moved "outside" the component
- Results hidden by default
- To vote, user clicks on desired option and then "cast vote"
- "Show results" sort of expands the component and reveals the results above the vote options. Users may collapse this module.
- Clicking on the amount of votes opens the votes modal → add a chevron to the right of the vote amount

Image

Image

andyesp:
Closes #806 

image

image

image


1emu (approved):

src/components/Icon/ChevronRight.tsx
ncomerci, line 1: TODO for another moment: I'm thinking if we could define a common interface for the parameters of the icons we use, because now each one gets more or less things depending on what is needed in each case

ncomerci (approved):

src/components/Icon/ChevronRight.tsx
andyesp: Yes! That's a nice refactor to do

PR #879, opened 2023-03-27
refactor: color variables

heyimjuani:
Includes some minor style corrections


src/components/Category/CategoryOption.css
```diff
@@ -31,6 +31,11 @@
   background-color: var(--background-disabled);
 }
 
+.CategoryOption.CategoryOption--active .dg.Paragraph.Paragraph--semi-bold,
+.FilterLabel--active .dg.Paragraph.Paragraph--tiny {
```
andyesp, line 35: Please move the FilterLabel change to FilterLabel.css

src/components/Proposal/ProposalItem.css
```diff
@@ -1,6 +1,7 @@
-.ProposalItem {
+.ProposalItem,
+.ui.card.Link.ProposalItem {
```
andyesp, line 2: You can remove this line if you add !important to the border in line 4.

src/theme.css
```diff
@@ -1,43 +1,93 @@
...
+  --yellow-200: rgba(254, 247, 210, 1);
+  --yellow-800: rgba(var(--yellow-raw), 1);
+
+  --fuchsia-200: rgba(255, 228, 255, 1);
+  --fuchsia-200t: rgba(var(--fuchsia-raw), 0.16);
+  --fuchsia-500t: rgba(var(--fuchsia-raw), 0.5);
+  --fuchsia-800: rgba(var(--fuchsia-raw), 1);
+
+  --green-200: rgba(225, 243, 214, 1);
+  --green-200t: rgba(var(--green-raw), 0.16);
```
andyesp, line 42: What's the `t` in the theme colors? 🤔

heyimjuani: Visually it would be the same color - except one uses relies on Alpha values and the other is solid (100% opacity)

heyimjuani: This is bound to change as soon as we implement some upcoming CSS color goodies and/or CSS preprocessors that would enable us to operate with colors in a more programatic way. Think expressions/functions like darken/lighten/mix.

1emu: > I don't like the abbreviation. Maybe `--transparent-green-200`... Why isn't it another color instead of a green-200 variation?

I also don't understand this, if we can have different shades of a color expressed as a numerical value: can the transparent ones be also a color between 0 and 999?

1emu: Take the reds for example:

```css
  --red-100: rgba(var(--red-raw), 0.16);
  --red-200: rgba(249, 214, 220, 1);
  --red-200t: rgba(var(--red-raw), 0.16);
```

We have `red-200t` that is the same as `red-100`. We could just use `red-100` right?

And in the same way, we could do that for the greens, instead of 

```css
  --green-200: rgba(225, 243, 214, 1);
  --green-200t: rgba(var(--green-raw), 0.16);
```

we could have 

```css
  --green-100: rgba(var(--green-raw), 0.16);
  --green-200: rgba(225, 243, 214, 1);
  --green-400: rgba(var(--green-raw), 0.4);
  --green-800: rgba(var(--green-raw), 1);
  --green-900: rgba(50, 133, 0, 1);
  ```
  
WDYT?

andyesp: We can also think the colors with alpha as another color palette.
I find useful to check what color is being used in the figma by inspecting the component and finding the color style name. Are we going to keep that 1:1 relationship with figma?

andyesp (approved):

1emu (approved):

PR #887, opened 2023-03-28
feat: Specific grant filters for proposals

### LINKED ISSUE #821: Proposals can be filtered by grant category
https://www.figma.com/file/e0IFJjplUdYdQJ2NtLwvMf/DAO-v2-(Guane)?node-id=9760%3A2660

Image

ncomerci:
This closes #821 

image


src/entities/Grant/types.ts
```diff
@@ -39,6 +40,13 @@ export enum NewGrantCategory {
   Sponsorship = 'Sponsorship',
 }
 
+export enum SubtypeAlternativeOptions {
+  All = 'all_grants',
+  Legacy = 'legacy',
+}
+
+export type SubtypeOptions = `${NewGrantCategory}` | `${SubtypeAlternativeOptions}`
```
andyesp, line 48: Maybe we can move this to proposal types and make this a `ProposalSubtype`? 🤔 Or `GrantSubtype`...

src/components/Category/CategoryOption.tsx
```diff
@@ -58,30 +74,100 @@ export default React.memo(function CategoryOption({ active, type, className, cou
...
+    if (subtype === SubtypeAlternativeOptions.All) {
+      url.searchParams.delete('subtype')
+    } else {
+      url.searchParams.set('subtype', subtype)
+    }
+    const newHref = url.pathname + '?' + url.searchParams.toString()
+    return newHref
+  }
+  const hasSubtypes = !!subtypes && subtypes.length > 0
+  const [isSubcategoriesOpen, setIsSubcategoriesOpen] = useState(false)
```
andyesp, line 89: `isSubtypesOpen`?

src/components/Icon/Arrow.tsx
```diff
@@ -5,11 +5,12 @@ import TokenList from 'decentraland-gatsby/dist/utils/dom/TokenList'
...
   filled: boolean
   className?: string
+  color?: string
 }
 
-function Arrow({ filled, className }: ArrowProps) {
-  const fill = filled ? 'var(--black-400)' : 'white'
-  const stroke = filled ? 'white' : 'var(--black-400)'
+function Arrow({ filled, className, color = 'var(--black-400)' }: ArrowProps) {
+  const fill = filled ? color : 'rgba(255, 255, 255, 0)'
```
andyesp, line 12: Is `rgba(255, 255, 255, 0` supposed to be transparent? Doesn't `transparent` work?

src/components/Search/CategoryFilter.tsx
```diff
@@ -63,6 +63,7 @@ export default React.memo(function CategoryFilter({
       )}
       {filters.map((value, index) => {
         const label = toSnakeCase(value)
+        const isGrant = value == ProposalType.Grant
```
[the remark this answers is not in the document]
1emu: Shouldn't you use `isGrantType` for this? 

I'm thinking that if the answer is no and that function is for something else (more specific), it could have a more specific name. Like `isGrantSubtype` maybe?

src/entities/Proposal/model.ts
```diff
@@ -235,120 +241,140 @@ export default class ProposalModel extends Model<ProposalAttributes> {
...
+      ${conditional(!!user && !coauthor, SQL`AND p."user" = ${user}`)} 
+      ${conditional(
+        !!coauthor && !!user,
+        SQL`AND lower(c."address") = lower(${user}) AND (CASE WHEN p."finish_at" < NOW() THEN c."status" IN (${CoauthorStatus.APPROVED}, ${CoauthorStatus.REJECTED}) ELSE TRUE END)`
+      )}
+      ${conditional(!!type, SQL`AND "type" = ${type}`)} 
       ${conditional(
-        !!filter.coauthor && !!filter.user,
-        SQL`AND lower(c."address") = lower(${filter.user}) AND (CASE WHEN p."finish_at" < NOW() THEN c."status" IN (${CoauthorStatus.APPROVED}, ${CoauthorStatus.REJECTED}) ELSE TRUE END)`
+        !!subtype,
```
1emu, line 265: It'd be good to extract this logic to a method:

```typescript
${conditional(!!subtype, SQL`AND (${this.getSubtypeQuery(subtype)})`)}
```

So you can use it in the other query of the Model as well.

```typescript
private static getSubtypeQuery(subtype: string) {
    return subtype === SubtypeAlternativeOptions.Legacy
      ? this.getLegacyGrantCategoryQuery()
      : SQL`p."configuration" @> '{"category": "${SQL.raw(subtype || '')}"}'`
  }
```

Also, since you are checking `!!subtype`, I think you don't need to
`SQL.raw(subtype || '')` and you can just `{SQL.raw(subtype)`

```typescript
private static getLegacyGrantCategoryQuery() {
    return join(
      Object.values(OldGrantCategory).map(
        (category) => SQL`p."configuration" @> '{"category": "${SQL.raw(category)}"}'`
      ),
      SQL` OR `
    )
  }
  ```

**Review by 1emu — APPROVED** [review]
Nice workaround to the proposal sub categories. LGTM, left some suggestions.

File: src/components/Category/CategoryOption.tsx [hunk 6]
```diff
@@ -42,8 +47,19 @@ export const getCategoryIcon = (type: string, variant?: CategoryIconVariant, siz
...
+  type,
+  className,
+  count,
+  subtypes,
+  ...props
+}: CategoryOptionProps) {
   const t = useFormatMessage()
+  const location = useLocation()
+  const params = useMemo(() => new URLSearchParams(location.search), [location.search])
+  const currentSubtype = useMemo(() => (params.get('subtype') as SubtypeOptions) || undefined, [params])
```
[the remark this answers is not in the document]
ncomerci: Yes! I saw it and I fixed it in the PR that comes after this one!

image
image

src/entities/Grant/types.ts
```diff
@@ -61,6 +69,17 @@ export enum GrantStatus {
   Revoked = 'Revoked',
 }
 
+export function isGrantType(value: string | null | undefined) {
```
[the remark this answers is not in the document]
ncomerci: What name do you suggest? Because the purpose of these functions is to determine if a string X is a grant category type

src/entities/Proposal/model.ts
ncomerci: It is true that I am checking `!!subtype` but that check is the p1 parameter of the `conditional(p1, p2)` function.  The function returns p2 if p1 is true, but whatever is used in p2 is evaluated regardless of whether p1 is true or not (it is not lazy like p1 && p2).

That said, the SQL.raw() function fails if the parameter passed to it is undefined.

src/entities/Grant/types.ts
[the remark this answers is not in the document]
1emu: Oh ok, if this is for general purpose then I'm ok with it

1emu (approved):

PR #913, opened 2023-04-04
feat: Hiring proposal

### LINKED ISSUE #853: New Category: Hiring Proposal
New Category: Committee Members
- Sub Category: Add Committee Member / Remove Committee Member
6M VP /  Duration: 4 Weeks
Min VP to submit: 2,500VP
Yes/No/Abstain

User will be able to select first the Add/Remove Committee Members and then select if Add or Remove.
Remove category will always be available
If there is no slot available in any committee, user's won't be able to select a committee in the selection dropdown.

Form Structure:
Committee (Select only between Committees with open slots)
Address
Reasons to add/remove
Evidence

Title structure
Should [XXXX](DCL NAME/AVATAR NAME/ADDRESS) be added to [Committee Name]?
Should [XXXX](DCL NAME/AVATAR NAME/ADDRESS) be removed to [Committee Name]?

- [ ] Check Committee availability in Team Transparency JSON

Transparency: JSON file with Committee, addresses associated with that committee and available slots.
- Dao Committee (3)
- Curations Committee (15)
- Revocations Committee (7)
- SAB (5)

@heyimjuani To define icon & color

Estimate: 6 working days

ncomerci:
This closes #853 

- [x] Disable 'Add' if there're no available positions
- [x] Disable 'committees' without available positions
- [x] Test max characters
- [x] Remove mock add


src/clients/DclData.ts
```diff
@@ -36,7 +44,7 @@ export type TransparencyData = {
     total: bigint
     budget: bigint
   }
-  teams: Team[]
+  teams: Committee[]
```
andyesp, line 47:
```suggestion
  commitees: Committee[]
```
Shouldn't we update Transparency so it is using the "commitee" name there instead of team?

src/components/Proposal/Submit/ProposalSubmitHiringPage.tsx
```diff
@@ -0,0 +1,292 @@
...
+import CommitteeMembersDropdown from './CommitteeMembersDropdown'
+import './ProposalSubmitHiringPage.css'
+
+interface Props {
+  type: HiringType
+  committees: CommitteeName[]
+  isCommitteesLoading?: boolean
+}
+
+type HiringState = {
```
andyesp, line 42: I think you can use a NewHiringType or Omit `type` from NewHiringType here

src/entities/Proposal/templates/hiring.ts
```diff
@@ -0,0 +1,45 @@
+import { addressShortener } from '../../../helpers'
+import { HiringType, NewProposalHiring } from '../types'
+
+import { template } from './utils'
+
+type HiringConfig = NewProposalHiring & { name: string }
```
andyesp, line 6: Isn't name part of the `NewProposalHiring`? 🤔  Or is this because it's probably undefined?

src/entities/Proposal/types.ts
```diff
@@ -126,6 +135,20 @@ export function getPoiTypeAction(poiType: PoiType) {
   return poiType.split('_')[0] // "add" | "remove"
 }
 
+export function toHiringType(value: string | null | undefined): HiringType | null {
```
andyesp, line 138: What about changing this to `isValidHiringType` and return a boolean accordingly??

andyesp: I'd also use the same logic in both Hiring and POI submit pages.

src/entities/Proposal/utils.ts
```diff
@@ -4,6 +4,7 @@ import Time from 'decentraland-gatsby/dist/utils/date/Time'
 import 'isomorphic-fetch'
 import numeral from 'numeral'
 
+import { DclData } from '../../clients/DclData'
```
andyesp, line 7: Is this being used?

src/components/Modal/AddDeleteProposalModal/AddDeleteProposalModal.tsx
```diff
@@ -4,30 +4,45 @@ import Paragraph from 'decentraland-gatsby/dist/components/Text/Paragraph'
...
 
-export function POIProposalModal({ ...props }) {
+export type AddDeleteProposalModalProps = ModalProps & {
+  title: 'poi' | 'hiring'
+  addType: PoiType | HiringType
+  addHref: string
+  isAddDisabled?: boolean
+  removeType: PoiType | HiringType
+  removeHref: string
+  isRemoveDisabled?: boolean
```
andyesp, line 20: `isDeleteDisabled`? Component has `Delete` in its name.

src/entities/Committee/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { Committee, CommitteeName, DclData } from '../../clients/DclData'
+
+export async function getCommitteesWithOpenSlots(): Promise<Committee[]> {
+  // const { teams: committees } = await DclData.get().getData()
```
1emu, line 4: Reminder to remove this mock

1emu: We should try/catch somewhere because the DclData call could fail. 
One implementation path could be to try/catch here, and return an empty list. But this would mean we'd have to try catch on other places that we use `getData()` too.

Maybe `getData()` could return a valid `null object` that has `{ teams: [] }` and any other expected data when the request fails.

src/entities/Proposal/templates/hiring.ts
1emu: Since we are doing the `getName` check, I don't think we need this

src/entities/Proposal/types.ts
1emu: I agree that returning boolean is better than returning null. 
But if you need to cast some value to a valid type, it's useful to have a fn that can do it.

A good approach IMHO is something like we have with proposal status:

```typescript
export function toHiringType(value: string | null | undefined, orElse: () => any): HiringType | any {
  return isValidHiringType(value) ? (value as HiringType) : orElse()
}
```

You can always cast something but it forces you to provide an explicit solution when the casting fails (which could be returning a default type, returning null, or throwing an error, reporting, etc).

```typescript 
toHiringType(something, () => null)
toHiringType(something, () => { throw new Error("Invalid hiring type" })
```

1emu (approved):
I'm wondering if we really need to type the valid Committees... 
Because adding one in transparency will mean that we need to add it here. What if we leave them as a string?

src/entities/Proposal/templates/hiring.ts
ncomerci: Yes, my bad! It's outdated

1emu (approved):

src/entities/Proposal/types.ts
```diff
@@ -126,6 +135,24 @@ export function getPoiTypeAction(poiType: PoiType) {
...
+    case HiringType.Add:
+    case HiringType.Remove:
+      return true
+    default:
+      return false
+  }
+}
+
+export function toHiringType<T>(value: string | null | undefined, orElse: () => T): HiringType | T {
+  return isHiringType(value) ? (value as HiringType) : orElse()
```
andyesp, line 149: Why don't just return `null` instead of orElse()?

src/components/Proposal/Submit/CommitteeMembersDropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+      }))
+    },
+    [committees, onOptionClick]
+  )
+
+  return (
+    <Controller
+      control={control}
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      render={({ field: { ref, onChange, ...field } }) => (
```
andyesp, line 46: is `onChange` on purpose here to avoid passing it to Dropdown?

andyesp (approved):

PR #937, opened 2023-04-19
feat: Link forum profile with DAO profile

ncomerci:
Closed because we took other approach

### DIFF
--- a/src/entities/Discourse/types.ts
+++ b/src/entities/Discourse/types.ts
@@ -1,3 +1,77 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 export type DiscourseConnetTokenBody = {
   payload: string
+}
+
+export type DiscourseAttributes = {
+  address: string
+  forum_id: number
+  user_api_key_encrypted: string
+  expiration_date: string
+}
+
+export interface DiscourseUser {
+  errors?: string[]
+  error_type?: string
+  current_user?: {
+    id: number
+    username: string
+    name: string
+    avatar_template: string
+    unread_notifications: number
+    unread_private_messages: number
+    unread_high_priority_notifications: number
+    all_unread_notifications_count: number
+    read_first_notification: boolean
+    admin: boolean
+    notification_channel_position: null
+    moderator: boolean
+    staff: boolean
+    whisperer: boolean
+    title: null
+    any_posts: boolean
+    trust_level: number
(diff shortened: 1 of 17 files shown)


src/entities/Discourse/types.ts
```diff
@@ -0,0 +1,77 @@
...
+  address: string
+  forum_id: number
+  user_api_key_encrypted: string
+  expiration_date: string
+}
+
+export interface DiscourseUser {
+  errors?: string[]
+  error_type?: string
+  current_user?: {
```
andyesp, line 16: Do we need to type the keys we don't use from the current_user object?

src/entities/Proposal/routes.ts
```diff
@@ -485,7 +486,25 @@ export async function proposalComments(req: Request<{ proposal: string }>) {
...
+      map[user.forum_id] = user.address
+      return map
+    }, {} as Record<number, string>)
+
+    filteredComments.comments = filteredComments.comments.map((comment) => {
+      comment.address = userMapping[comment.id]
+      return comment
+    })
+
+    return filteredComments
```
andyesp, line 507: I prefer here to do something like
```
return {
  ...filteredComments,
  comments: filteredComments.comments.map((comment) => ({ ...comment, address: userMapping[comment.id] }))
}

File: src/services/DiscourseConnect.ts [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+  getUrl() {
+    return this.url
+  }
+
+  getUserApiKey(encodedKey: string): string {
+    const result = decrypt(encodedKey, this.privateKey)
+    const parsedResult = JSON.parse(result)
+
+    if (parsedResult.nonce !== this.nonce) {
+      throw new Error('Nonce does not match')
```
andyesp, line 46: Return forbidden or unauthorized?

src/services/DiscourseConnect.ts
```diff
@@ -0,0 +1,51 @@
+import { hostname } from 'os'
+
+import { FORUM_URL, GOVERNANCE_API } from '../constants'
+import { decrypt, generateAsymmetricKeys } from '../helpers'
+
+const APP_NAME = GOVERNANCE_API.replace(/\/api/, '')
```
andyesp, line 6: We are doing this in `server.ts` too. Can we move it to a helper file?

PR #951, opened 2023-04-25
feat: Added payment token in grants view

### LINKED ISSUE #935: Add payment token to proposal detail view
We could add it to grant size

Grant size: $40.000 USD (in MANA) / (in DAI).

ncomerci:
This closes #935 

image

### Legacy Grants
image

### DIFF
--- a/src/components/Proposal/View/Categories/GrantProposalView.tsx
+++ b/src/components/Proposal/View/Categories/GrantProposalView.tsx
@@ -4,6 +4,7 @@ import useFormatMessage, { useIntl } from 'decentraland-gatsby/dist/hooks/useFor
 
 import { NewGrantCategory } from '../../../../entities/Grant/types'
 import { GrantProposalConfiguration } from '../../../../entities/Proposal/types'
+import { CURRENCY_FORMAT_OPTIONS } from '../../../../helpers'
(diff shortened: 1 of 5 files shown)


src/components/Proposal/View/Categories/GrantProposalView.tsx
```diff
@@ -41,7 +43,14 @@ function GrantProposalView({ config }: Props) {
       <ProposalDescriptionItem title={t('page.proposal_view.grant.abstract_title')} body={abstract} />
       <ProposalDescriptionItem
         title={t('page.proposal_view.grant.size_title')}
-        body={`${intl.formatNumber(size)} USD`}
+        body={
+          paymentToken
+            ? t('page.proposal_view.grant.size_body', {
+                amount: intl.formatNumber(size, CURRENCY_FORMAT_OPTIONS as any),
```
andyesp, line 49: You can move `intl.formatNumber(...)` to a variable to be reused.
`const formattedAmount = intl.formatNumber(size, CURRENCY_FORMAT_OPTIONS as any)`

(Currency can also be applied to the formatter so it returns the value with the USD string, but we can check that later)

andyesp (approved):

1emu (approved):

PR #956, opened 2023-04-26
feat: Remove leading option on active proposals for logged in users

### LINKED ISSUE #954: Remove leading option on active proposals in the proposal view for logged in users

ncomerci:
This closes #954 

Logged out: See the results
Logged in and voted: See the results
Logged in and not voted: Hide the results

### Logged in
image

### Logged out
image

### DIFF
--- a/src/components/Status/LeadingOption.tsx
+++ b/src/components/Status/LeadingOption.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo } from 'react'
 
+import useAuthContext from 'decentraland-gatsby/dist/context/Auth/useAuthContext'
 import useFormatMessage from 'decentraland-gatsby/dist/hooks/useFormatMessage'
 
 import { ProposalStatus } from '../../entities/Proposal/types'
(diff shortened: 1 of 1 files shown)


src/components/Status/LeadingOption.tsx
```diff
@@ -15,6 +16,9 @@ export type LeadingOptionProps = React.HTMLAttributes<HTMLDivElement> & {
 
 export default React.memo(function LeadingOption({ status, leadingOption, metVP, userChoice }: LeadingOptionProps) {
   const t = useFormatMessage()
+  const [user] = useAuthContext()
+
+  const hideLeadingOption = useMemo(() => !!user && !userChoice, [user, userChoice])
```
andyesp, line 21: I'd change this to `showLeadingOption` as I always find using the opposite confusing :P 
Also... I don't think is necessary to memoize this condition.

andyesp: You can also group all the conditions
`const showLeadingOption = status === ProposalStatus.Active && !!user && !userChoice`

ncomerci: At first I thought so but in this particular case, the 'show' has more than one condition. Instead, the 'hide' only has one condition and I felt it was easier to define it that way.

Regarding the Memo, it is true, it can be omitted :)

1emu (approved):

PR #983, opened 2023-05-18
feat: use react query to fetch profile, grants and transparency data

andyesp:
Reduces number of requests while navigating the app.


ncomerci (approved):
The ratio Impact/effort is over the clouds, I love it! 

I would suggest having a file where we have defined in one place all the keys with their respective stale times 😄

1emu (commented):
🔥

ncomerci (approved):

PR #990, opened 2023-05-22
feat: remove back button from proposal

andyesp:


src/pages/proposal.css
```diff
@@ -13,6 +35,7 @@
 }
 
 .ProposalDetailPage .ui.header {
+  margin-top: 0;
```
1emu, line 38: How about a 0.2 | 0.5rem top margin? 👀

1emu (approved):

PR #995, opened 2023-05-30
feat: tender updates

andyesp:
image

image

image


.env.example
```diff
@@ -57,6 +57,8 @@ GATSBY_DURATION_GOVERNANCE=1200
 GATSBY_DURATION_PITCH=1200
 GATSBY_DURATION_TENDER=1200
 GATSBY_DURATION_HIRING=1200
+DURATION_TENDER=1200
```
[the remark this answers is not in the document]
andyesp: It's backend only.

src/entities/Proposal/jobs.ts
```diff
@@ -97,6 +93,35 @@ function updateCategoryBudget(proposal: ProposalWithOutcome, budgetForProposal:
...
+      continue
+    }
+
+    pendingProposalsWithOutcome.push({ ...proposal, ...outcome })
+  }
+
+  return pendingProposalsWithOutcome
+}
+
+async function getFinishedTenderProposals(pendingProposals: ProposalAttributes[]) {
```
1emu, line 111: This is weird because it's called `getFinishedTenderProposals` but returns pending tender proposals.
I'd try to avoid using pending if it's going to be a proposal status.

How about getFinishableTenderProposals? Or something that indicates that they are proposals that need to be processed.

src/entities/Proposal/jobs.ts
```diff
@@ -107,40 +132,51 @@ async function categorizeProposals(
...
+        finishedProposals.push(proposal)
         break
       case ProposalOutcome.ACCEPTED:
-        if (proposalWithOutcome.type !== ProposalType.Grant || proposalWithOutcome.start_at < BUDGETING_START_DATE) {
-          acceptedProposals.push(proposalWithOutcome)
-          break
+        if (proposal.type !== ProposalType.Grant && proposal.type !== ProposalType.Tender) {
+          acceptedProposals.push(proposal)
+        } else if (proposal.type === ProposalType.Tender) {
+          const tenderProposals = pendingProposalsWithOutcome.filter(
```
1emu, line 153: Yo could extract this to 

```typescript
function getWinnerTender(pendingProposalsWithOutcome: any[], proposal: ProposalWithOutcome) {
  const tenderProposals = pendingProposalsWithOutcome.filter(
    (item) =>
      item.type === ProposalType.Tender &&
      item.configuration.linked_proposal_id === proposal.configuration.linked_proposal_id
  )
  return orderBy(tenderProposals, 'winnerVotingPower', 'desc')[0]
}
```

```
else if (proposal.type === ProposalType.Tender) {
          const winnerTenderProposal = getWinnerTender(pendingProposalsWithOutcome, proposal)
          if (winnerTenderProposal.id === proposal.id) {
            acceptedProposals.push(proposal)
          } else {
            rejectedProposals.push(proposal)
          }
        }
        ```

File: src/entities/Proposal/routes.ts [hunk 4]
```diff
@@ -404,18 +404,39 @@ export async function createProposalTender(req: WithAuth) {
...
   await validateSubmissionThreshold(user, SUBMISSION_THRESHOLD_TENDER)
 
+  const tenderProposals = await ProposalModel.getProposalList({
+    linkedProposalId: configuration.linked_proposal_id,
+    order: 'ASC',
+  })
+
+  if (
+    tenderProposals.find(
+      (proposal) => proposal.status === ProposalStatus.Passed || proposal.status === ProposalStatus.Rejected
```
1emu, line 417: This is interesting. Is this the only way we have today to know if a pitch proposal finished the tendering process?

src/entities/Proposal/routes.ts
```diff
@@ -404,18 +404,39 @@ export async function createProposalTender(req: WithAuth) {
...
+  ) {
+    throw new RequestError('Pitch proposal already went through the tender process')
+  }
+
+  const hasStartedTenderProcess = Time(tenderProposals[0].start_at).isBefore(Time())
+  if (hasStartedTenderProcess) {
+    throw new RequestError('Tender process already started for this pitch proposal')
+  }
+
+  const start_at =
```
1emu, line 428: This for me is a hint that we might be needing a new abstraction for this... 

For example, a `TenderingProcess` object, with a `start_at`, and `end_at`, that we can associate to a pitch proposal and/or it's tenders, and use it to make this decisions (instead of having to use each tender, and having to get the associated tenders list)

1emu (approved):

src/components/Proposal/View/CompetingTenders.tsx
```diff
@@ -0,0 +1,51 @@
...
+import ProposalCard from './ProposalCard'
+
+interface Props {
+  proposal: ProposalAttributes
+}
+
+export default function CompetingTenders({ proposal }: Props) {
+  const t = useFormatMessage()
+  const [isSiderbarOpen, setIsSidebarOpen] = useState(false)
+  const { tenderProposals } = useTenderProposals(proposal.configuration.linked_proposal_id, proposal?.type)
```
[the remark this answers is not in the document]
andyesp: I'll update the hook to only fetch proposals if there is a `linked_proposal_id`. We don't need a check here.

src/components/Status/LeadingOption.tsx
```diff
@@ -7,53 +7,53 @@ import { ProposalStatus } from '../../entities/Proposal/types'
...
 
   const proposalFinished = useMemo(() => {
-    return [
-      ProposalStatus.Passed,
-      ProposalStatus.Rejected,
-      ProposalStatus.Finished,
-      ProposalStatus.OutOfBudget,
-      ProposalStatus.Enacted,
-    ].includes(status)
+    return FINISHED_STATUSES.includes(status)
```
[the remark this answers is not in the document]
andyesp: I wouldn't move this to a helper if we are not reusing it anywhere else. But in this case, as it can make the platform more "understandable" we could do that. What could be a good name for this? As calling it "finished" doesn't seem right having a "Finished" status.

src/entities/Proposal/routes.ts
andyesp: Yea... We could have something in the pitch configuration? Like a Tender status. Or something in the middle between two pitch/tender proposals.

.env.example
```diff
@@ -57,6 +57,8 @@ GATSBY_DURATION_GOVERNANCE=1200
...
+START_DATE_GAP_DURATION_TENDER=1200
```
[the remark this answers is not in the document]
1emu: `TENDER_SUBMISSION_WINDOW_DURATION` ?

src/components/Proposal/View/ProposalPromotionSection.tsx
```diff
@@ -68,19 +70,17 @@ export default function ProposalPromotionSection({ proposal, loading }: Props) {
     }
   }
 
+  const isPromoteDisabled =
+    (type === ProposalType.Pitch && (hasTenderProcessFinished || hasTenderProcessStarted)) ||
```
1emu, line 74: If `hasTenderProcessFinished` is true, then `hasTenderProcessStarted` should always be true right? 

So this could be 

```typescript
const isPromoteDisabled =
    (type === ProposalType.Pitch && hasTenderProcessStarted) || type === ProposalType.Tender
```

Unless `hasTenderProcessStarted` means that the tender process is ongoing, and could be false once it's finished, in which case I'd call it `isTenderProcessOngoing` or open.

src/components/Proposal/View/ProposalVoting/ProposalVotingSection.tsx
```diff
@@ -98,6 +96,7 @@ const ProposalVotingSection = ({
 
   const proposalVotingSectionLoading = loading || accountState.loading || delegationState.loading || isLoadingVp
   const showGetInvolvedQuestion = !!proposal && !proposalVotingSectionLoading && !hasVoted && !finished
+  const isPendingStatus = proposal?.status === ProposalStatus.Pending
```
1emu, line 99: `isProposalPending`?

src/entities/Proposal/jobs.ts
```diff
@@ -22,6 +22,13 @@ type Outcome = {
 
 type ProposalWithOutcome = ProposalAttributes & Outcome
 
+export async function activateProposals(context: JobContext) {
+  const activatedProposals = await ProposalModel.activateProposals()
+  if (activatedProposals > 0) {
```
1emu, line 27: This check 🔥🔥🔥🔥🔥!!!!

1emu (approved):
send it fuse

src/components/Proposal/View/ProposalPromotionSection.tsx
andyesp: You are absolutely right

1emu (approved):

ncomerci (approved):

PR #1005, opened 2023-06-05
Competing proposals v2

1emu:
Update:

Added separation between proposal bars
image
image

Some cases to see how it looks:

# Overbudget cases:

When the displayed initiative asked for a relatively low budget (and it's on the very edge of the bar)
image

Different initiative for same category
image

When there is no granted funds for the category, but requested budget already is over the available:
image

Proposal popups have a max width of 150px
image

image

The downside of the "This initiative" popup always on is that sometimes you won't be able to read the legends. I think it'd be better to hide it when you hover over a competing proposal.
image

# Within budget cases

image

image

image

Not "unselecting" the initiative under review also means that selected competing proposals bars need to be on top of it when hovered
image


src/components/Proposal/View/Budget/ContestedBudgetCard.css
```diff
@@ -0,0 +1,98 @@
...
+
+.GrantedFundsBarOverbudget {
+  height: 8px !important;
+}
+
+.CompetingProposalBudgetBar {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  /*border-radius: 4px;*/
```
andyesp, line 34: 👀

src/components/Proposal/View/Budget/ContestedBudgetSubLabel.css
```diff
@@ -0,0 +1,28 @@
+.ContestedBudgetSubLabel {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  font-family: var(--font-family);
```
andyesp, line 5: Where is this var defined?

andyesp (approved):

src/components/Proposal/View/Budget/ContestedBudgetSubLabel.css
1emu: decentraland-ui > styles.css

andyesp (approved):

ncomerci (approved):

andyesp (approved):

PR #1033, opened 2023-06-20
feat: use sublabel component in all forms

andyesp:


src/components/Common/SubLabel.css
```diff
@@ -7,6 +7,6 @@
   font-style: normal;
   font-weight: 400;
   font-size: 15px;
```
1emu, line 9: I think we can remove font-weight, style, color, and size if we are using Text's css

1emu (approved):

src/components/Common/SubLabel.css
andyesp: This is overriding the Markdown component styles 😞

PR #1051, opened 2023-06-27
refactor: Replace useClipboardCopy from dcl-gatsby

### LINKED ISSUE #1050: replace useClipboardCopy from dcl-gatsby

ncomerci:
This closes #1050


1emu (approved):
I have some doubts about the naming

```
 return {
    copiedValue,
    copy,
    clear,
  }
```
  
The word `copy` could also refer to the copied value, so in some contexts it might not be entirely clear 
(when we do `const { copy } = useClipboardCopy(Time.Second)` for example)

I think `copyValue` could be a slightly better name for the function, but I'm not against leaving it as it is.

andyesp (approved):

PR #1063, opened 2023-07-03
refactor: Replace profile loader from dcl-gatsby

### LINKED ISSUE #1045: Replace profile loader from dcl-gatsby

ncomerci:
This closes #1045


src/pages/proposal.css
```diff
@@ -22,10 +22,6 @@
   min-height: 24px;
 }
 
-.ProposalDetailPage__Description img {
```
1emu, line 25: Are you sure this is not being used? (like, an img inside a proposal markdown)

src/services/DiscordService.ts
```diff
@@ -127,16 +127,15 @@ export class DiscordService {
 
     if (user) {
       try {
-        const profile = await profiles.load(user)
-        const hasDclProfile = !!profile && !profile.isDefaultProfile
-        const profileHasName = hasDclProfile && !!profile.name && profile.name.length > 0
+        const profile = await getProfile(user)
+        const profileHasName = !!profile && !!profile.name && profile.name.length > 0
```
1emu, line 131: shouldn't you also check that it's not the default profile? Or use `hasDclProfile` instead

`const profileHasName = hasDclProfile && !!profile.name && profile.name.length > 0`

Which is the same we are doing in the hook, and makes me think it could be part of the return variables

1emu (approved):

src/components/Common/Avatar.tsx
```diff
@@ -18,42 +19,27 @@ export enum AvatarSize {
...
-      onError={() => setFailed(true)}
-      className={classNames(
-        'Avatar',
-        `Avatar--${size || AvatarSize.Mini}`,
-        `Avatar--${((address || '')[2] || '').toLowerCase()}`,
-        !src && isLoadingProfile && `Avatar--loading`,
-        className
-      )}
-    />
+  const getClassNames = () =>
```
andyesp, line 31: I think you can write this 
```
const avatarClassNames = classNames(
      'Avatar',
      `Avatar--${size || AvatarSize.Mini}`,
      `Avatar--${((address || '')[2] || '').toLowerCase()}`,
      isLoadingProfile && `Avatar--loading`,
      className
    )
```

src/entities/Proposal/routes.ts
```diff
@@ -315,8 +315,12 @@ export async function createProposalHiring(req: WithAuth) {
   }
 
   if (!configuration.name) {
-    const profile = await profiles.load(configuration.address)
-    configuration.name = profile?.name || undefined
+    try {
+      const profile = await getProfile(configuration.address)
+      configuration.name = profile?.name || undefined
```
andyesp, line 320: Undefined not necessary here. `configuration.name = profile?.name`

src/pages/proposal.css
ncomerci: Yeap, my bad. Idk why I deleted this. Thanks for the heads up!

src/services/DiscordService.ts
ncomerci: isDefaultProfile was a dcl-gatsby add-on. In this case it is not necessary!

1emu (approved):

PR #1066, opened 2023-07-03
fix: hiring type and constant name

andyesp:


src/entities/Proposal/types.ts
```diff
@@ -632,7 +632,6 @@ export const ProposalRequiredVP = {
   [ProposalType.Governance]: requiredVotingPower(VOTING_POWER_TO_PASS_GOVERNANCE, 0),
   [ProposalType.Pitch]: requiredVotingPower(VOTING_POWER_TO_PASS_PITCH, 0),
   [ProposalType.Tender]: requiredVotingPower(VOTING_POWER_TO_PASS_TENDER, 0),
-  [ProposalType.Hiring]: requiredVotingPower(VOTING_POWER_TO_PASS_HIRING, 0),
```
1emu, line 635: But what about `required_to_pass: ProposalRequiredVP[ProposalType.Hiring]` ? 
It does not have a default case

1emu (approved):

src/entities/Proposal/constants.ts
ncomerci, line 1: I think this change also affects the definitions repo, doesn't it?

andyesp: No, it is ok in the definitions repo. And that variable is only used in the frontend.

ncomerci (approved):

PR #1102, opened 2023-07-18
feat: finish bid proposals

### LINKED ISSUE #1101: Finish bid proposals
Calculate winning bid. Check tender calculation. It should work the same.

andyesp:
Closes #1101

### DIFF
--- a/src/entities/Proposal/jobs.ts
+++ b/src/entities/Proposal/jobs.ts
@@ -9,7 +9,7 @@ import { ErrorCategory } from '../../utils/errorCategories'
 import { Budget } from '../Budget/types'
 
 import ProposalModel from './model'
-import { ProposalOutcome, ProposalWithOutcome, calculateOutcome, getWinnerTender } from './outcome'
+import { ProposalOutcome, ProposalWithOutcome, calculateOutcome, getWinnerBiddingAndTenderingProposal } from './outcome'
 import { commentProposalUpdateInDiscourse } from './routes'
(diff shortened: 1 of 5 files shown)


src/entities/Proposal/jobs.ts
```diff
@@ -107,18 +107,25 @@ async function getProposalsWithOutcome(proposals: ProposalAttributes[], context:
...
-  if (pendingTenderProposals.length > 0) {
-    const linkedProposalIds = [...new Set(pendingTenderProposals.map((item) => item.configuration.linked_proposal_id))]
-    pendingTenderProposals = []
+export async function getFinishabledLinkedProposals(
+  pendingProposals: ProposalAttributes[],
+  type: ProposalType.Bid | ProposalType.Tender
+) {
+  let proposals = pendingProposals.filter((item) => item.type === type)
+  if (proposals.length > 0) {
+    const linkedProposalIds = [...new Set(proposals.map((item) => item.configuration.linked_proposal_id))]
```
ncomerci, line 116: Why a Set is needed here? Is it possible to have a single "linked_proposal_id" more than once?

1emu (approved):

src/entities/Proposal/jobs.ts
andyesp: Yes. This is to get all the linked proposal ids so we can query the db for those proposals.
We could just grab the proposals from the "finishable proposals query" but this was added just to make sure we are getting them all.

ncomerci (approved):

PR #1113, opened 2023-07-20
chore: get addresses votes

1emu:


andyesp (approved):

src/services/SnapshotService.ts
```diff
@@ -97,4 +98,11 @@ export class SnapshotService {
     }
     return { snapshotStatus, snapshotSpace }
   }
+
+  static async getAddressesVotes(addresses: string[], first?: number, skip?: number) {
+    if (isNumber(first) && isNumber(skip)) {
```
andyesp, line 103: I think you also have native options like Number.isInteger

1emu: Yes but the compiler was less happy with it

PR #1117, opened 2023-07-21
chore: snapshot proposal votes

1emu:


src/back/routes/snapshot.ts
```diff
@@ -5,16 +5,25 @@ import { Request } from 'express'
...
 
 async function getAddressesVotes(req: Request<{ spaceName?: string }>) {
   const { addresses } = req.body
   return await SnapshotService.getAddressesVotes(addresses)
 }
+
+async function getProposalVotes(req: Request<{ id?: string }>) {
+  const { id } = req.params
+  if (!id || id.length === 0) {
+    return []
```
andyesp, line 26: What if we throw a 500 error here?

andyesp (approved):

ncomerci (approved):

src/back/routes/snapshot.ts
1emu: Will do in next snapshot PR

PR #1134, opened 2023-07-25
chore: snapshot api key pt3

1emu:


ncomerci (approved):

src/back/routes/validations.ts
```diff
@@ -0,0 +1,56 @@
...
+    'strategies',
+    'discussion',
+    'plugins',
+  ]
+  if (!fields.every((field) => validFields.includes(field))) {
+    throw new RequestError('Invalid fields', RequestError.BadRequest)
+  }
+}
+
+export function validateProposalId(id?: string, optional: 'optional' | 'required' = 'required') {
```
andyesp, line 46: The argument is called `optional` but you can pass `require` there? 🤔 

I'd change the second argument to something like... an options object `{ required: boolean }` or a boolean argument `required = true`. 

Also I think the validation is missing the case where the value is not optional and the id.length is 0.

src/clients/Governance.ts
```diff
@@ -481,9 +481,42 @@ export class Governance extends API {
...
+  async getPendingProposals(start: Date, end: Date, fields: (keyof SnapshotProposal)[], limit: number) {
+    const response = await this.fetch<ApiResponse<Partial<SnapshotProposal>[]>>(
+      `/snapshot/proposals/pending`,
+      this.options().method('POST').json({ start, end, fields, limit })
+    )
+    return response.data
+  }
+
+  async getVpDistribution(address: string, proposalSnapshotId?: string) {
+    let url = `/snapshot/vp-distribution/${address}`
```
andyesp, line 515:
```
cosnt snapshotId = proposalSnapshotId ? `/${proposalSnapshotId}` : ''
const url = `/snapshot/vp-distribution/${address}${proposalSnapshotId}`
```

src/clients/Governance.ts
```diff
@@ -481,9 +481,42 @@ export class Governance extends API {
...
+
+  async getSnapshotProposals(start: Date, end: Date, fields: (keyof SnapshotProposal)[]) {
+    const response = await this.fetch<ApiResponse<Partial<SnapshotProposal>[]>>(
+      `/snapshot/proposals`,
+      this.options().method('POST').json({ start, end, fields })
+    )
+    return response.data
+  }
+
+  async getPendingProposals(start: Date, end: Date, fields: (keyof SnapshotProposal)[], limit: number) {
```
andyesp, line 506: Maybe it would be great if this receives an object. It will make it easier to read everywhere we call for this method.

E.g:
```
Governance.get().getPendingProposals({
  start,
  date,
  fields: ['id', 'total'],
  limit: 5
})
```

andyesp (approved):

src/back/routes/validations.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { validateProposalId } from './validations'
+
+describe('validateProposalId', () => {
+  describe('when id is optional', () => {
+    it('should not throw an error when not provided', () => {
+      expect(() => validateProposalId(undefined, 'optional')).not.toThrow()
+    })
+
+    it('should not throw an error when it is an empty string', () => {
```
andyesp, line 11: `should throw` here

src/back/routes/validations.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    })
+
+    it('should not throw an error for a valid proposal id', () => {
+      expect(() => validateProposalId('any string is fine really', 'optional')).not.toThrow()
+    })
+  })
+
+  describe('when it is required', () => {
+    it('should not throw an error for a valid proposal id', () => {
+      expect(() => validateProposalId('any string is fine really')).not.toThrow()
```
andyesp, line 26: This is misleading as this string is not a valid proposal Id. Should we use `isUUID` in function to validate?

src/entities/Proposal/routes.ts
```diff
@@ -9,6 +9,7 @@ import { Request } from 'express'
 import isEthereumAddress from 'validator/lib/isEthereumAddress'
 import isUUID from 'validator/lib/isUUID'
 
+import { validateAddress } from '../../back/routes/validations'
```
andyesp, line 12: Maybe move this to another folder? `back/helpers/validations`, `back/utils/validations`

1emu: I'm not sure about this one. 

My logic was that if we are going to have all routes in `back/routes`, and this validations are only going to be used in there (because they validate route params and and throw RequestErrors), then it's ok to have them in there.
I'm open to having them somewhere else, as long as we don't use them to validate outside routes.

WDYT?

PR #1223, opened 2023-08-30
fix: Badge airdrops in batches

ncomerci:


1emu (approved):

src/back/jobs/BadgeAirdrop.ts
```diff
@@ -28,5 +30,7 @@ async function runQueuedAirdropJobs() {
 }
 
 async function giveAndRevokeLandOwnerBadges() {
-  await BadgesService.giveAndRevokeLandOwnerBadges()
+  if (config.getEnv() === Env.PRODUCTION) {
```
andyesp, line 33: `config` is frontend only 🤔

1emu: can we use isProdEnv()?

andyesp: Yes!

src/entities/Snapshot/utils.ts
```diff
@@ -203,3 +204,8 @@ export function isSameAddress(userAddress?: string | null, address?: string | nu
     getChecksumAddress(userAddress) === getChecksumAddress(address)
   )
 }
+
+export async function getUsersWhoVoted(usersWithoutBadge: string[]) {
```
andyesp, line 208: Maybe `getUsersFromVotes`? And rename `usersWithoutBadge` to `users`? Or move this to Badge utils if we want to make it specific to badges.

andyesp (approved):

PR #1224, opened 2023-08-30
feat: top voter badge

### LINKED ISSUE #1128: Top Voter Badge
The Top Voter badge is granted to users who were the ones that casted more votes in the DAO at any given month
The Top Voter badge will be airdropped to users who voted with +5VP in the last month (If any of their votes in the target month were with less than 5VP, that shouldn't count)
The Top Voter will be granted the first calendar month of the following month after the votes were casted
The Top Voter badge will either: Be disabled after the following calendar month after the votes were casted or stay active and another Badge will be created for the next month
The Top Voter badge should state in the metadata the month when the votes were casted

1emu:
Closes #1128


src/back/jobs/BadgeAirdrop.ts
```diff
@@ -33,3 +33,8 @@ async function giveAndRevokeLandOwnerBadges() {
     await BadgesService.giveAndRevokeLandOwnerBadges()
   }
 }
+
+export async function giveTopVoterBadges() {
```
ncomerci, line 37: Do we want this to run also in testing env or only in prod?

ncomerci (approved):

src/back/jobs/BadgeAirdrop.ts
1emu: Good question. I think we do want it to run in testing env, since I expect for us to be the top voters on our space.

src/back/utils/contractInteractions.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+  await txn.wait()
+  logger.log('Airdropped badge with txn hash:', txn.hash)
+  return txn.hash
+}
+
+export async function reinstateBadge(badgeId: string) {
+  const provider = RpcService.getPolygonProvider()
+  const raftOwner = new ethers.Wallet(RAFT_OWNER_PK, provider)
+  const contract = new ethers.Contract(POLYGON_BADGES_CONTRACT_ADDRESS, BadgesAbi, raftOwner)
```
andyesp, line 50: We could extract this first three lines to a `getBadgesContract` function or something like that right? Maybe we could have two. One to get the raftOwner and another one to get the contract as I see we are repeating these lines in all these functions.

andyesp (approved):

PR #1266, opened 2023-09-15
Feat/add new user menu

flobarreto:


package.json
```diff
@@ -84,6 +85,7 @@
     "gatsby-plugin-typescript": "^4.10.0",
     "gatsby-source-filesystem": "^4.10.0",
     "gatsby-transformer-sharp": "^4.10.0",
+    "isomorphic-fetch": "^3.0.0",
```
1emu, line 88: I think you don't need this dep :)

andyesp: I think we are using it from the `decentraland-gatsby` dependencies. So I think it is ok to add it to our package json since we are importing and using it in some files.

ncomerci (approved):

PR #1283, opened 2023-09-21
fix: show/hide proposal vp chart

ncomerci:


src/pages/proposal.tsx
```diff
@@ -170,6 +170,7 @@ export default function ProposalPage() {
   const { votes, isLoadingVotes, reloadVotes } = useProposalVotes(proposal?.id)
   const { highQualityVotes, lowQualityVotes } = useMemo(() => getVoteSegmentation(votes), [votes])
   const { surveyTopics, isLoadingSurveyTopics } = useSurveyTopics(proposal?.id)
+  const [showResults, setShowResults] = useState(false)
```
1emu, line 173: I think you can just add this to the proposalPageState, that is already being shared with the ProposalGovernanceSection

ncomerci: Great!

1emu (approved):

PR #1323, opened 2023-10-06
fix: snapshot status banner

### LINKED ISSUE #1318: fix snapshot status banner

1emu:
Closes #1318 

Switched to error rate monitoring:

image

image

image


src/constants.ts
```diff
@@ -38,4 +38,7 @@ export const DEBUG_ADDRESSES = (process.env.DEBUG_ADDRESSES || '')
   .split(',')
   .filter(isEthereumAddress)
   .map((address) => address.toLowerCase())
-export const SNAPSHOT_STATUS_ENABLED = false
+
+export const SNAPSHOT_STATUS_ENABLED = process.env.SNAPSHOT_STATUS_ENABLED === 'true'
```
andyesp, line 42: You can use `getBooleanStringVar`

src/services/SnapshotStatusService.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (errorRate > SNAPSHOT_STATUS_ERROR_RATE_THRESHOLD) {
+      health = ServiceHealth.Failing
+    }
+    return { errorRate, health }
+  }
+
+  private static visualizeRequestResults(requestResults: boolean[]) {
+    const numRows = 1
+    const numCols = requestResults.length
+    const greenPoint = chalk.green('●')
```
ncomerci, line 79: Should we keep printing this?

ncomerci (approved):

src/services/SnapshotStatusService.ts
1emu: I don't see why not :) 
It will only appear in the logs when there are errors in the buffer, and it's useful to see how the errors are mixed with the OK requests

src/constants.ts
1emu: I thought about it but it uses a clientEnv method that takes the var from config.json, I wasn't sure that I could use it in this case

andyesp: We can check this in another PR

andyesp (approved):

src/constants.ts
1emu: https://github.com/decentraland/governance/issues/1328

PR #1335, opened 2023-10-10
chore: add pull request template

### LINKED ISSUE #1375: Add pull request template to repository

andyesp:
Closes #1375

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,7 +6,7 @@ E.g: *Closes/Fixes #1234*
 - [ ]  **Functionality validation**
     - [ ]  Feature working as described in the related issue
     - [ ]  New environmental variables have been set
-    - [ ]  New services methods are properly tested
+    - [ ]  New API services methods have tests
     - [ ]  Code errors are being logged in the code health service
 - [ ]  **Product validation**
     - [ ]  UI changes look the same as in Figma
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,13 +3,10 @@
 **Related Issue**
 E.g: *Closes/Fixes #1234*
 
-- [ ]  **Code Quality**
-    - [ ]  The code follows established coding standards and style guidelines
-    - [ ]  The code is properly commented on and provides context when needed (complex feature or hard to understand)
 - [ ]  **Functionality validation**
     - [ ]  Feature working as described in the related issue
     - [ ]  New environmental variables have been set
-    - [ ]  New functions are properly tested (When should we add tests?)
+    - [ ]  New services methods are properly tested
     - [ ]  Code errors are being logged in the code health service
 - [ ]  **Product validation**
     - [ ]  UI changes look the same as in Figma
@@ -18,6 +15,6 @@ E.g: *Closes/Fixes #1234*
     - [ ]  Preview link if it’s a customer facing issue
 - [ ] **Documentation**
     - [ ] README updated
-    - [ ] New API changes are properly documented
+    - [ ] New API changes are documented
 - [ ] **Accessibility**
     - [ ] UI changes are accessible
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,8 +8,7 @@ E.g: *Closes/Fixes #1234*
     - [ ]  The code is properly commented on and provides context when needed (complex feature or hard to understand)
 - [ ]  **Functionality validation**
     - [ ]  Feature working as described in the related issue
-    - [ ]  Does this PR need environment variables to be set? If yes, have they been set?
-    - [ ]  New API endpoints are properly documented
+    - [ ]  New environmental variables have been set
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+**Screenshots**
+
+**Related Issue**
+E.g: *Closes/Fixes #1234*
+
+- [ ]  **Code Quality**
+    - [ ]  The code follows established coding standards and style guidelines
```
1emu, line 7: It'd be nice to have a link to some guidelines. 
I think foundation had one we could use.

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+    - [ ]  The code is properly commented on and provides context when needed (complex feature or hard to understand)
+- [ ]  **Functionality validation**
+    - [ ]  Feature working as described in the related issue
+    - [ ]  Does this PR need environment variables to be set? If yes, have they been set?
```
1emu, line 11: How about linking the definitions PR from the governance PRs if there is one?

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+    - [ ]  New API endpoints are properly documented
+    - [ ]  New functions are properly tested (When should we add tests?)
```
1emu, line 13: I think we could aim to have all service methods tested as the baseline. Routes and jobs on most cases should rely on a service that should be tested.

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+    - [ ]  Code errors are being logged in the code health service
+- [ ]  **Product validation**
+    - [ ]  UI changes look the same as in Figma
+    - [ ]  Usage events are being sent to the product analytics service
```
1emu, line 17: The usage events we want to register should be defined at the issue.

It would be nice to have an Issue definition checklist too!

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+    - [ ]  New common components have been documented in Storybook
```
1emu, line 18: I worry this might add too much overhead to a feature branch. 

Creating a component and adding it to the storybook, and checking that the colors are defined in figma, etc could be a whole issue/PR in itself. This could also make us more aware of what components we decide to create, outside of what we already have.

This could be part of the issues checklist: If a new component is needed, check it has a ticket specifically for that.

1emu (changes requested):
I like this checklist. 

I would also add the following:

```md
- [ ] **Security**
    - [ ] No sensitive data hardcoded
    - [ ] Proper input validation and sanitation (check routes and db queries)
    - [ ] Auth checks for new endpoints
- [ ] **Performance**
    - [ ] No N+1 queries
    - [ ] Proper use of caching/stale times
- [ ] **Rollback Plan**
    - [ ] Database migrations are reversible
    - [ ] Feature flags used for risky changes
- [ ] **Documentation**
    - [ ] README updated
    - [ ] Public API changes are documented (remove this one from Functionality Validation)
- [ ] **Accessibility**
    - [ ] UI changes are accessible, i.e., screen reader tests, Guane's magic check
```

Also, one important thing we are missing and we should add to one of this sections is the **mobile view** of any new page/component we add.
That could also be added to the Issue/Feature Checklist

.github/pull_request_template.md
ginocingolani: We have some guidelines enforced by eslint, but I agree it would be good to have those standards and guidelines published somewhere

.github/pull_request_template.md
ginocingolani: Agree with this and will work on it. But besides having the definition at the issue level, this checklist should validate that the changes were actually made.

.github/pull_request_template.md
andyesp: I'm not sure about Foundation guidelines but we could find one that fits our coding style and conventions.

.github/pull_request_template.md
andyesp: It's the same overhead as creating a test for a new service in the backend.

.github/pull_request_template.md
1emu: Yeah, maybe the when and where for the FF should come with the issue definition, and here we just check if it's included in the PR

ginocingolani: Until we agree on a coding standard, let's remove this

.github/pull_request_template.md
ginocingolani: https://github.com/decentraland/governance/issues/1372 WIP

.github/pull_request_template.md
ginocingolani: my opinion on this is that if you treat design system issues as separate tasks, they never get done. Let's have a reusable mindset when building UIs and try to build custom components as much as possible. In case we have something specific and common that we want to add, let's estimate the effort when estimating the task.

1emu (approved):

PR #1368, opened 2023-10-25
fix: proposal render issue

andyesp:


src/components/Proposal/View/ProposalCard.tsx
```diff
@@ -75,7 +75,7 @@ export default function ProposalCard({
                 <span>{' · '}</span>
               </>
             )}
-            <span className="ProposalCard__DetailsItem ProposalCard__DetailsOnlyDesktop">
+            <span className="ProposalCard__DetailsItem">
```
1emu, line 78: Should we use NotMobile here?

src/pages/proposal.tsx
```diff
@@ -7,7 +7,6 @@ import Head from 'decentraland-gatsby/dist/components/Head/Head'
 import NotFound from 'decentraland-gatsby/dist/components/Layout/NotFound'
 import useAuthContext from 'decentraland-gatsby/dist/context/Auth/useAuthContext'
 import usePatchState from 'decentraland-gatsby/dist/hooks/usePatchState'
-import { Loader } from 'decentraland-ui/dist/components/Loader/Loader'
 import { NotMobile, useMobileMediaQuery } from 'decentraland-ui/dist/components/Media/Media'
```
1emu, line 10: If dcl-ui components are not working for us, maybe we should stop using them. WDYT?

1emu (approved):
🕵️

PR #1398, opened 2023-11-09
feat: change last notification endpoint return

andyesp:


ncomerci (approved):

src/components/Notifications/Notifications.tsx
```diff
@@ -93,27 +93,15 @@ export default function Notifications() {
...
     enabled: !!user && isSubscribed,
-    retry: 3,
   })
 
   const hasNewNotifications =
-    Number(userNotifications?.length) > 0 &&
-    !!lastNotificationId &&
-    latestNotification &&
-    latestNotification !== Number(lastNotificationId)
+    Number(userNotifications?.length) > 0 && latestNotification && latestNotification !== Number(lastNotificationId)
```
1emu, line 104: Are this cast to Number necessary?

1emu (approved):

src/components/Notifications/Notifications.tsx
andyesp: Converted here as those values could be null or undefined and typescript didn't like to use the > operator with non-number values. But I'll remove the second one as it isn't needed.

PR #1405, opened 2023-11-16
chore: update push sdk to support sepolia

andyesp:


src/entities/SurveyTopic/decoder.test.ts
```diff
@@ -14,44 +12,41 @@ const SENTIMENT_SURVEY: TopicFeedback[] = [
 ]
 
 describe('decode', () => {
-  def('decodedSurvey', () => decodeSurvey(get.encodedSurvey))
+  const decodedSurvey = (encodedSurvey: Record<string, unknown>) => decodeSurvey(encodedSurvey)
 
   describe('and empty survey', () => {
-    def('encodedSurvey', () => '')
+    // TODO: encodedSurvey shouldn't be a string. Maybe { survey: [] }?
```
1emu, line 18: I'm not sure about this.

The survey encodes to a string because we are saving a stringifyed version of the survey in snapshot's `reason` field. 

I don't remember the details, but there could be old proposals with an empty reason, and I think snapshot returns an empty or null value.

src/entities/SurveyTopic/decoder.test.ts
```diff
@@ -65,91 +60,84 @@ describe('decode', () => {
...
-      expect(get.decodedSurvey).toEqual([])
+      expect(decodedSurvey(encodedSurvey)).toEqual([])
     })
   })
 
   describe('a survey encoded as a string', () => {
-    def('encodedSurvey', () => {
-      return `{"survey":[{ topic_id: '12345', reaction: 'neutral' }, { topic_id: '22222' , reaction: 'love' }, { topic_id: '33333', reaction: 'empty' }]}`
-    })
+    // TODO: Is this a valid test if typescript is already enforcing what type the function receives?
```
1emu, line 118: I think we can delete this one!

src/entities/SurveyTopic/decoder.test.ts
1emu: Also, it's possible to send a custom or empty reason voting directly from snapshot, so we have no guarantee about what we are going to receive

1emu (approved):

src/entities/SurveyTopic/decoder.test.ts
andyesp: Then we should update decodeSurvey params to be `encodedSurvey?: string | Record<string, unknown>`.
I'd remove the TODO. We could review this another time.

ncomerci (approved):

PR #1462, opened 2023-12-13
chore: catalyst profiles refactor

1emu:
**Screenshots**

image

image

image


src/components/Profile/VotedProposalsBox.tsx
```diff
@@ -25,8 +25,8 @@ function VotedProposalsBox({ address }: Props) {
       {isLoading && <SkeletonBars amount={votes.length || 3} height={83.5} />}
       {!isLoading &&
         (votes.length > 0 ? (
-          votes.map((vote) => {
-            return <ProfileProposalItem key={vote.id} votedProposal={vote} />
+          votes.map((vote, idx) => {
+            return <ProfileProposalItem key={vote.id + idx} votedProposal={vote} />
```
andyesp, line 29: Why vote.id isn't enough?

src/hooks/useProfiles.ts
```diff
@@ -1,34 +1,24 @@
...
-
     try {
-      const profiles = await getProfiles(validAddresses)
-      validAddressesProfiles = profiles.map<CatalystProfileStatus>((profile, idx) => ({
-        profile: profile || createDefaultAvatar(validAddresses[idx]),
-        isDefaultProfile: !profile,
-      }))
+      return await getProfiles(validAddresses)
     } catch (error) {
+      //TODO: should we report this?
```
andyesp, line 18: I wouldn't report this as it could be an error from the catalyst instead.

src/utils/Catalyst/index.ts
```diff
@@ -3,36 +3,32 @@ import isEthereumAddress from 'validator/lib/isEthereumAddress'
...
 import { isSameAddress } from '../../entities/Snapshot/utils'
 
-import { CatalystProfile, DisplayableNameAndAvatar, ProfileResponse } from './types'
+import { CatalystProfile, DclProfile, ProfileResponse } from './types'
 
 const CATALYST_URL = 'https://peer.decentraland.org'
 export const DEFAULT_AVATAR_IMAGE = 'https://decentraland.org/images/male.png'
 
-export function getDisplayableUsername(profile: CatalystProfile | null, user: string) {
+export function getUsernameAndAvatar(profile: CatalystProfile | null, address: string): DclProfile {
```
andyesp, line 11: This function is not used anywhere else, export can be removed.
I'd rename this to `getDclProfile` or `generateDclProfile`.

src/components/Common/AvatarFace.tsx
```diff
@@ -0,0 +1,27 @@
...
+import './AvatarFace.css'
+
+export type AvatarFaceProps = {
+  avatar?: string
+  size?: 'tiny' | 'small' | 'medium' | 'large' | 'responsive'
+  inline?: boolean
+  className?: string
+}
+
+export class AvatarFace extends React.PureComponent<AvatarFaceProps> {
```
andyesp, line 12: Could you write this as a function? It's a small refactor 😄

src/components/Common/AvatarFace.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
+
+import './AvatarFace.css'
+
+export type AvatarFaceProps = {
```
andyesp, line 5: Export not needed.

src/components/Common/AvatarFace.tsx
andyesp: Also why copying this component from dcl/ui is necessary?

src/components/Profile/VotedProposalsBox.tsx
1emu: I thought the same, but saw an error of elements with repeated id

image

src/components/Common/Avatar.tsx
```diff
@@ -22,21 +20,23 @@ type Props = {
   address?: string
   className?: string
   src?: string
+  avatar?: string
+  isLoadingDclProfile?: boolean
 }
 
-export default function Avatar({ address, size, src, className }: Props) {
-  const { profile, isLoadingProfile } = useProfile(address, !!src)
+// TODO: avatar should be mandatory when no src is provided
```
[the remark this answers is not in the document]
1emu: `avatar` is the flat img of the user's in-world avatar (in case the user has one), or an image of a default avatar when user has no catalyst profile.

`src` is a link to an image (any image) which we are using for the forum avatars. 

So when the user has no custom avatar, we try to use the forum one.

```typescript
    {hasCustomAvatar ? (
            <Avatar address={address} avatar={avatar} size="medium" isLoadingDclProfile={isLoadingDclProfile} />
          ) : (
            <Avatar size="medium" src={avatarUrl} />
          )}
          ```
          
I think we could refactor this component in this PR :)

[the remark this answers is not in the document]
Comment (andyesp) [reply src/components/Common/Avatar.tsx]: AvatarFace is rendering an img too, even though it is adding some classes. Couldn't we just use AvatarFace in Avatar and remove the condition where an image is rendered?

File: src/components/Common/Avatar.tsx [hunk 7]
```diff
@@ -22,21 +22,24 @@ type Props = {
...
     className
   )
 
   return src ? (
     <img src={src} className={avatarClassNames} />
+  ) : avatar ? (
+    <AvatarFace avatar={avatar} className={avatarClassNames} />
   ) : (
-    <AvatarFace avatar={profile} className={avatarClassNames} />
+    <AvatarFace avatar={DEFAULT_AVATAR_IMAGE} className={avatarClassNames} />
```
andyesp, line 43: `avatar || DEFAULT_AVATAR_IMAGE`

andyesp (approved):

PR #1473, opened 2023-12-20
fix: urls formed by get url filters function

andyesp:


src/helpers/index.ts
```diff
@@ -53,7 +55,8 @@ export function getUrlFilters<T>(filterKey: string, params: URLSearchParams, val
     newParams.delete('subtype')
   }
   const stringParams = newParams.toString()
-  return `${location.pathname}${stringParams === '' ? '' : '?' + stringParams}`
+  const pathname = location.pathname.replace('/governance', '')
```
1emu, line 58: I wonder if we need to do something similar in the usePreventNavigation hook

1emu (approved):

PR #1474, opened 2023-12-20
feat: Financial submit cards

ncomerci:


src/back/routes/vestings.ts
```diff
@@ -7,9 +7,14 @@ import { VestingService } from '../../services/VestingService'
 import { validateAddress } from '../utils/validations'
 
 export default routes((router) => {
+  router.get('/all-vestings', handleAPI(getAllVestings))
```
andyesp, line 10: Just `vestings`? We don't have `all-projects` or `all-proposals`

src/components/Updates/FinancialCard.css
```diff
@@ -0,0 +1,24 @@
+.ui.card.FinancialCard {
+  margin: 0;
+  width: 100%;
+  padding: 16px;
+  background-color: var(--black-700);
+  box-shadow: 0px 1px 8px 0px var(--alpha-black-200);
+}
+
+.ui.card.FinancialCard .Text {
```
andyesp, line 9: Please use a classname to modify the Text component 🙏

src/components/Updates/FinancialCard.css
```diff
@@ -0,0 +1,24 @@
+.ui.card.FinancialCard {
```
andyesp, line 1: I think you can avoid using `.ui.card` with some `!important`s. That is the way 🙏

src/components/Updates/FinancialCard.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.FinancialCard__Value {
+  display: flex;
+  align-items: center;
+  margin: 8px 0;
+  gap: 10px;
+}
+
+.FinancialCard__Value > .Text {
```
andyesp, line 22: Here too.

src/back/routes/vestings.ts
ncomerci: Will be removed in the next PR

PR #1523, opened 2024-01-05
feat: Navbar2 integration

braianj:
Integration of the new [Navbar2](https://ui.decentraland.org/?path=/story/navbar2--with-notification-pending)


braianj:
change made in #1525 1525

PR #1532, opened 2024-01-10
fix: disable notifications when pk is not present

1emu:
Notifications enabled and no PK (server does not crash)
image

Notifications disabled
image

Notifications enabled and PK
image

### DIFF
--- a/src/back/services/notification.ts
+++ b/src/back/services/notification.ts
@@ -31,7 +31,6 @@ const PUSH_CHANNEL_OWNER_PK = process.env.PUSH_CHANNEL_OWNER_PK
 const PUSH_API_URL = process.env.PUSH_API_URL
 
 function getSigner() {
-  console.log('notificationsServiceEnabled', NOTIFICATIONS_SERVICE_ENABLED)
   if (NOTIFICATIONS_SERVICE_ENABLED) {
     const hexPk = `0x${PUSH_CHANNEL_OWNER_PK}`
     if (!PUSH_CHANNEL_OWNER_PK || !ethers.utils.isHexString(hexPk, 32)) {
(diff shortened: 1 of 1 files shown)


src/back/services/notification.ts
```diff
@@ -29,22 +29,38 @@ import PushAPI = require('@pushprotocol/restapi')
 const chainId = isProdEnv() ? ChainId.ETHEREUM_MAINNET : ChainId.ETHEREUM_SEPOLIA
 const PUSH_CHANNEL_OWNER_PK = process.env.PUSH_CHANNEL_OWNER_PK
 const PUSH_API_URL = process.env.PUSH_API_URL
-const pkAddress = `0x${PUSH_CHANNEL_OWNER_PK}`
-const signer = NOTIFICATIONS_SERVICE_ENABLED ? new ethers.Wallet(pkAddress) : undefined
+
+function getSigner() {
+  console.log('notificationsServiceEnabled', NOTIFICATIONS_SERVICE_ENABLED)
```
andyesp, line 34: 👀 I'd remove this console.log

src/back/services/notification.ts
```diff
@@ -29,22 +29,38 @@ import PushAPI = require('@pushprotocol/restapi')
...
+  if (NOTIFICATIONS_SERVICE_ENABLED) {
```
andyesp, line 35: I'd return 
```
if (!NOTIFICATIONS_SERVICE_ENABLED) {
  return undefined
}
```
first. So there is no if inside if. And you are also checking for the "negative" value first in the next condition.

ncomerci (approved):

andyesp (approved):

PR #1552, opened 2024-01-15
feat: discourse comments on activity ticker

1emu:
Part 1 of #1535 

Uses new env var: `DISCOURSE_WEBHOOK_SECRET`

image

image

image


src/back/routes/events.ts
```diff
@@ -23,3 +25,11 @@ async function voted(req: WithAuth) {
   validateRequiredString('choice', req.body.choice)
   return await EventsService.voted(req.body.proposalId, req.body.proposalTitle, req.body.choice, user)
 }
+
+async function newDiscourseEvent(req: Request) {
```
andyesp, line 29: `commented`

src/back/utils/validations.ts
```diff
@@ -102,3 +106,19 @@ export function validateDebugAddress(user: string | undefined) {
...
+    const payload = req.body
+    const calculatedSignature = 'sha256='.concat(
+      crypto.createHmac('sha256', DISCOURSE_WEBHOOK_SECRET).update(JSON.stringify(payload)).digest('hex')
+    )
+
+    if (providedSignature !== calculatedSignature) {
+      throw new RequestError('Invalid signature', RequestError.Forbidden)
+    }
+  } else {
+    logger.warning('Unable to find secret for discourse webhook validation')
```
andyesp, line 122: I think we should also throw an error here so it doesn't keep the execution going wherever this is used 🤔

src/back/services/events.ts
```diff
@@ -152,4 +159,58 @@ export class EventsService {
 
     return profiles
   }
+
+  static async commented(eventId: string | undefined, event: string | undefined, discoursePost: DiscourseWebhookPost) {
```
andyesp, line 163: Shouldn't `eventId` be `discourseEventId`?

src/back/services/events.ts
```diff
@@ -152,4 +159,58 @@ export class EventsService {
...
+    try {
+      if (
+        !eventId ||
+        !event ||
+        event !== 'post_created' || //TODO: we don't register editions or deletions
+        (await EventModel.alreadyRegistered(eventId)) ||
```
andyesp, line 169: Maybe rename to `isCommentAlreadyRegistered`?

src/back/routes/events.ts
```diff
@@ -23,3 +25,11 @@ async function voted(req: WithAuth) {
...
+  validateDiscourseWebhookSignature(req)
+
+  const eventId = req.get('X-Discourse-Event-Id')
+  const event = req.get('X-Discourse-Event')
```
andyesp, line 33: Should we validate here if these values are present before calling the service?

src/back/services/events.ts
```diff
@@ -152,4 +159,58 @@ export class EventsService {
...
+        address: user.address,
+        event_type: EventType.Commented,
+        event_data: {
+          discourse_event_id: eventId,
+          discourse_event: event,
+          discourse_post: discoursePost,
+          proposal_id: commentedProposal.id,
+          proposal_title: commentedProposal.title,
+        },
+        created_at: new Date(), //TODO: should we use the discord event creation date?
```
andyesp, line 208: If we have it in the discoursePost object let's do it. Maybe we should do that for votes and proposal too 🤔

andyesp: Not sure now... haha

src/back/models/Event.ts
```diff
@@ -1,18 +1,20 @@
...
 
-import { Event } from '../../shared/types/events'
+import { Event, EventType } from '../../shared/types/events'
 
 export default class EventModel extends Model<Event> {
   static tableName = 'events'
   static withTimestamps = false
   static primaryKey = 'id'
 
+  // TODO: do we want to show comment events on the activity ticker?
```
andyesp, line 11: Yes!!!

src/back/services/events.ts
1emu: How about `isDiscourseEventRegistered`?  Cause I'm not certain we are not going to register other discord events (like topic creations, or editions/deletions), and they all share this event id

ncomerci (approved):

PR #1557, opened 2024-01-16
refactor: remove some decentraland-gatsby dependencies

andyesp:


1emu (approved):

1emu (approved):

src/utils/logger.ts
```diff
@@ -0,0 +1,43 @@
...
+  warn(message: string, data: LoggerData = {}): void {
+    return this.write(message, { level: 'warning', data })
+  }
+
+  error(message: string, data: LoggerData = {}): void {
+    if (data instanceof Error) {
+      data = {
+        ...data,
+        message: data.message,
+        stack: data.stack,
```
1emu, line 35: this makes me think we could include the error stack property in the ErrorService reports

1emu (approved):

1emu (approved):

src/utils/logger.ts
andyesp: Yeah... We also repeat this in another function... we could unify this logic.

ncomerci (approved):

PR #1567, opened 2024-01-18
chore: change badges image source

### LINKED ISSUE #1562: Get badges images from other source than IPFS

1emu:
Closes #1562 

Badges that match a known github badge use github link

image

Badges that don't, use the ipfs as default (and are reported)

image

image


src/services/BadgesService.ts
```diff
@@ -310,4 +313,36 @@ export class BadgesService {
...
+  }
+
+  static toGovernanceBadgeSpec(otterspaceBadges: OtterspaceBadge[]): GovernanceBadgeSpec {
+    if (otterspaceBadges.length === 0) throw new Error('No badges found')
+    const { name, description, image } = otterspaceBadges[0].spec.metadata
+    const badges = otterspaceBadges.map(BadgesService.toGovernanceBadge)
+    return { name, description, image: BadgesService.getGithubImagePermalink(name, image), badges }
+  }
+
+  static getGithubImagePermalink(badgeName: string, ipfsLink: string): string {
```
andyesp, line 340: This method doesn't need to be static right? It is only used here so it can be private. And you can call it `this.getGithubImagePermalink`?

src/services/BadgesService.ts
```diff
@@ -310,4 +313,36 @@ export class BadgesService {
...
+    const badge = UPLOADED_BADGES.find((uploadedBadge) => badgeName.includes(uploadedBadge.name))
+    if (!badge) {
+      ErrorService.report('Could not find image for badge.', { badgeName, ipfsLink, category: ErrorCategory.Badges })
+      return getIpfsHttpsLink(ipfsLink)
+    }
+    return `${GITHUB_BADGE_IMAGE_BASE_URL}${badge.imageName}?raw=true`
```
andyesp, line 346: Maybe a `getGithubBadgeImageBaseUrl`?
```
const getGithubBadgeImageBaseUrl = imageName => `https://github.blabla${imageName}?raw=true`
```

andyesp (approved):

ncomerci (approved):

PR #1568, opened 2024-01-18
feat: New all budgets endpoint

### LINKED ISSUE #1565: New endpoint returning all budgets

ncomerci:
**Screenshots**
image

**Related Issue**
This closes #1565 

- [ ]  **Functionality validation**
    - [ ]  Feature working as described in the related issue
    - [ ]  New environment variables have been set
    - [ ]  New API services methods have tests
    - [ ]  Code errors are being logged in the code health service
- [ ]  **Product validation**
    - [ ]  UI changes look the same as in Figma
    - [ ]  Usage events are being sent to the product analytics service
    - [ ]  New common components have been documented in Storybook
    - [ ]  Preview link if it’s a customer facing issue
- [ ] **Documentation**
    - [ ] README updated
    - [ ] New API changes are documented
- [ ] **Accessibility**
    - [ ] UI changes are accessible

### DIFF
--- a/src/entities/QuarterBudget/model.ts
+++ b/src/entities/QuarterBudget/model.ts
@@ -211,7 +211,7 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
   }
 
   static parseBudgets(results: BudgetQueryResult[]): Budget[] {
-    const budgetsByQuarter = new Map<string, BudgetQueryResult[]>()
+    const budgetsByQuarter = new Map<BudgetQueryResult['id'], BudgetQueryResult[]>()
 
     results.forEach((result) => {
       const { id } = result
--- a/src/entities/QuarterBudget/model.ts
+++ b/src/entities/QuarterBudget/model.ts
@@ -91,6 +91,20 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
     }
   }
 
+  static async getAllBudgets(): Promise<Budget[]> {
+    const query = SQL`
+        SELECT qb.id, qcb.category, qcb.total as category_total, qcb.allocated as category_allocated, qb.start_at, qb.finish_at, qb.total
+        FROM ${table(QuarterCategoryBudgetModel)} as qcb 
+        INNER JOIN ${table(QuarterBudgetModel)} as qb
+        ON qcb.quarter_budget_id = qb.id
+    `
+    const result = await this.namedQuery('get_current_budget', query)
+    if (!result || result.length === 0) {
+      return []
+    }
+    return this.parseBudget(result)
+  }
+
   static async getCurrentBudget(): Promise<Budget | null> {
     const now = new Date()
     const query = SQL`
@@ -107,7 +121,7 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
     if (!result || result.length === 0) {
       return null
     }
-    return this.parseBudget(result)
+    return this.parseBudget(result)[0]
   }
 
   static async getBudgetForDate(dateWithinBudget: Date): Promise<Budget | null> {
@@ -130,7 +144,7 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
     if (!result || result.length === 0) {
       return null
     }
-    return this.parseBudget(result)
+    return this.parseBudget(result)[0]
   }
 
   static async getCategoryBudgetForCurrentQuarter(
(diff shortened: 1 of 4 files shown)


src/entities/QuarterBudget/model.ts
```diff
@@ -172,27 +186,43 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
     })
   }
 
-  static parseBudget(result: BudgetQueryResult[]): Budget {
-    const categoriesResults: Record<string, CategoryBudget> = {}
+  static parseBudget(result: BudgetQueryResult[]): Budget[] {
```
1emu, line 189: I'm finding this harder to follow than the previous `parseBudget` function.

How about having a `parseBudgets` function that groups the category budgets by id, and then calls the previous `parseBudget` on each group?


```typescript
static parseBudgets(results: BudgetQueryResult[]): Budget[] {
    const budgetsByQuarter = this.groupBudgetResultsByQuarter(results)
    const budgets: Budget[] = []

    budgetsByQuarter.forEach((resultsForQuarter) => {
      const budget = this.parseBudget(resultsForQuarter)
      budgets.push(budget)
    })

    return budgets
  }

  static groupBudgetResultsByQuarter(budgetResults: BudgetQueryResult[]): Map<string, BudgetQueryResult[]> {
    const groupedResults = new Map<string, BudgetQueryResult[]>()

    budgetResults.forEach((result) => {
      const { id } = result
      const group = groupedResults.get(id)

      if (group) {
        group.push(result)
      } else {
        groupedResults.set(id, [result])
      }
    })

    return groupedResults
  }
  ```
  
  Then you just use `return this.parseBudgets(results)`
  
  And you could leave the parseBudget in the other functions

Comment (andyesp) [reply src/entities/QuarterBudget/model.ts:189]: Agree parseBudget should only return one budget.

Comment (andyesp) [reply src/entities/QuarterBudget/model.ts:189]: Not sure about having two more functions only used on getAllBudgets though. I'll try to gather all that info in one function and see if it is understandable.

File: src/entities/QuarterBudget/model.ts [hunk 2]
```diff
@@ -195,4 +209,25 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
       categories: categoriesResults,
     }
   }
+
+  static parseBudgets(results: BudgetQueryResult[]): Budget[] {
+    const budgetsByQuarter = new Map<string, BudgetQueryResult[]>()
```
andyesp, line 214: Is `string` `BudgetQueryResult['id']`?

src/entities/QuarterBudget/model.ts
```diff
@@ -195,4 +209,25 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
...
+
+    results.forEach((result) => {
+      const { id } = result
+      const group = budgetsByQuarter.get(id)
+
+      if (group) {
```
andyesp, line 220: Couldn't you write here... `if (budgetsByQuarter.has(id)) {`?

src/entities/QuarterBudget/model.ts
```diff
@@ -195,4 +209,25 @@ export default class QuarterBudgetModel extends Model<QuarterBudgetAttributes> {
...
+        group.push(result)
+      } else {
+        budgetsByQuarter.set(id, [result])
+      }
+    })
+
+    const budgets: Budget[] = []
+
+    budgetsByQuarter.forEach((resultsForQuarter) => budgets.push(this.parseBudget(resultsForQuarter)))
```
andyesp, line 229: Can't this be done with a map function?
`const budgets = budgetsByQuarter.map(resultsForQuarter => this.parseBudget(resultsForQuarter)`?

src/entities/QuarterBudget/model.ts
ncomerci: Yes! But I'll change it to `BudgetQueryResult['id']`

src/entities/QuarterBudget/model.ts
ncomerci: The `Map` class does not contain the map() function so it would have to be transformed first to an Array to do that

It would be something like: `Array.from(budgetsByQuarter.values()).map(...)`

src/services/BudgetService.ts
```diff
@@ -125,6 +124,14 @@ export class BudgetService {
     return await QuarterBudgetModel.createNewBudgets(transparencyBudgets)
   }
 
+  static async getAllBudgets(): Promise<Budget[]> {
+    const budgets = await QuarterBudgetModel.getAllBudgets()
+    if (budgets.length === 0) {
+      ErrorService.report('Could not find any budgets', { category: ErrorCategory.Budget })
```
1emu, line 130: I'm not sure this is an error we should report... 🤔

1emu (approved):

andyesp (approved):

PR #1606, opened 2024-01-30
fix: delegation events dates

1emu:
This PR requires a change in `ops-param-governance-alchemy-delegations-webhook-secret` for prd and zone

Events are now saved in order of tx confirmation.

Creation date is almost the same for events of the same transaction, with a ms difference according to the log index. There shouldn't be more than two events for each tx log (clear/set).

Also, there shouldn't be a problem if there are two tx confirmed in the same block (though it would be very unlikely). 
Worst case scenario, we have 2 `clear` events with same created_at and 2 `set` events with same created_at, but different tx hashes and for different addresses, so it would read:

```md
X set delegation to A
Z set delegation to B
X removed delegation from Y
Z removed delegation from W
```

image

image

### DIFF
--- a/src/entities/Proposal/jobs.ts
+++ b/src/entities/Proposal/jobs.ts
@@ -172,12 +172,21 @@ async function prepareProposalsAndBudgetsUpdates(
   }
 }
 
+function logFinishableProposals(finishableProposals: ProposalAttributes[]) {
+  if (isProdEnv()) {
+    logger.log(`Finishable proposals:`)
+    finishableProposals.forEach((proposal, index) => {
+      console.log(`${index}: ${proposal.id} - ${proposal.title}`)
+    })
+  }
+}
+
 export async function finishProposal() {
   logger.log(`Running finish proposal job...`)
 
   try {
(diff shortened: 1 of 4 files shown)


ncomerci (approved):

src/entities/Proposal/jobs.ts
```diff
@@ -172,12 +172,21 @@ async function prepareProposalsAndBudgetsUpdates(
...
+  }
+}
+
 export async function finishProposal() {
   logger.log(`Running finish proposal job...`)
 
   try {
     const finishableProposals = await ProposalModel.getFinishableProposals()
-    logger.log(`Finishable proposals`, finishableProposals)
+    logFinishableProposals(finishableProposals)
```
andyesp, line 189: I think we can remove this log.

andyesp (approved):

src/entities/Proposal/jobs.ts
1emu: Didn't we looked for a finished proposal recently? (the one that never finished). Maybe it's still useful 🤷

PR #1701, opened 2024-03-05
chore: check for undisclosed funds before validating on front

1emu:
- [x] check for undisclosed funds before validating on front
- [x] check for undisclosed funds before validating on back
- [x] allow for updates without financial records

It might be out of scope for this issue, but I don't think we are validating that released funds match financial records in the back!


src/back/routes/update.ts
```diff
@@ -117,27 +120,60 @@ async function createProposalUpdate(req: WithAuth<Request<{ proposal: string }>>
...
+    VestingService.getVestingInfo(proposal.vesting_addresses),
+    UpdateService.getAllByProposalId(proposal.id),
+  ])
+
+  const releases = vestingData ? getReleases(vestingData) : undefined
+  const publicUpdates = getPublicUpdates(updates)
+  const latestUpdate = getLatestUpdate(publicUpdates || [])
+  const { releasedFunds } = getFundsReleasedSinceLatestUpdate(latestUpdate, releases)
+  return releasedFunds > 0 ? parseFinancialRecords(financial_records) : null //TODO: we should also validate that financial records match release funds?
+}
```
ncomerci, line 140: Not necessarily, it is allowed to disclose less than release

ncomerci (approved):

andyesp (approved):

PR #1747, opened 2024-04-03
fix: duplicate bids publication

### LINKED ISSUE #1746: Duplicated bids published

1emu:
Closes #1746

### DIFF
--- a/src/back/jobs/jobLocks.ts
+++ b/src/back/jobs/jobLocks.ts
@@ -1,0 +1,25 @@
+export const JOB_LOCKS = new Map<string, boolean>() // in-memory lock store
+
+const acquireLock = (jobName: string): boolean => {
+  if (JOB_LOCKS.get(jobName)) {
+    return false
+  }
+  JOB_LOCKS.set(jobName, true)
+  return true
+}
+
+const releaseLock = (jobName: string): void => {
+  JOB_LOCKS.delete(jobName)
+}
+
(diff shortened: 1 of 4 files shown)


src/back/jobs/jobLocks.ts
```diff
@@ -0,0 +1,29 @@
...
+  JOB_LOCKS.set(jobName, true)
+  return true
+}
+
+const releaseLock = (jobName: string): void => {
+  JOB_LOCKS.delete(jobName)
+}
+
+export const withLock = (jobName: string, jobFunction: JobFunction) => async (context: JobContext) => {
+  if (!acquireLock(jobName)) {
```
andyesp, line 20: I find confusing that this function returns a boolean.
Why don't you add another one to check if has acquired the lock?
And the acquireLock one only do the acquire thing.

src/server.ts
```diff
@@ -62,7 +63,7 @@ import { GOVERNANCE_URL } from './constants'
 const jobs = manager()
 jobs.cron('@eachMinute', finishProposal)
 jobs.cron('@eachMinute', activateProposals)
-jobs.cron('@eachMinute', publishBids)
+jobs.cron('@eachMinute', withLock('publishBids', publishBids))
```
ncomerci, line 66: Would it not be beneficial to also increase the frequency with which it runs? Maybe every 5 minutes for example

ncomerci (approved):

andyesp (approved):

andyesp (approved):

Issue #1772, opened 2024-04-29
Passed Grant/Bid creates new Project


andyesp:
Finish proposal job calls ProjectService.create with passed proposal.
Initial Project values:
* title: proposal title
* status: pending
* links: proposal links
* about: proposal about

New `project` entity:
* proposal_id
* title
* status (Pending, In Progress, Finished, Paused, Revoked)
* links
* about
* about_updated_by
* about_updated_date

New `project_personnel` entity: (Use case: we could show user participation in projects on the profile)
* project_id
* address?
* name
* role
* description
* link
* updated_by
* updated_date
* status (deleted or unassigned | assigned)

New `project_milestone` entity
* project_id
* title
* description
* status (pending | in progress | done)
* created_by
* created_date
* updated_by
* updated_date

New `project_milestone_update` entity
* update_id
* milestone_id
* description
* tasks

> What should the status of a project with a revoked vesting be? Should we add a "cancelled" status?

Let's use the same status we have right now, Pending, In Progress, Finished, Paused, Revoked

> Is project title the proposal title?

Yes, for now let's use the proposal title

> How are we going to calculate the "ongoing" date? Is it from the vesting contracts? Is it added manually from the GSS?

We should consider an ongoing project from the moment the vesting contract was initiated.
Project changes status from `Pending` to `In Progress` on proposal enactment.

andyesp:
- What should be the status of a project with a revoked vesting?

andyesp:
![Image](https://github.com/decentraland/governance/assets/8242162/3b65c423-cfde-4941-a3bc-2e7a1dfb79b0)
* Is project title the proposal title?
* How are we going to calculate the "ongoing" date? Is it from the vesting contracts? Is it added manually from the GSS?

PR #1783, opened 2024-05-07
chore: project entities

1emu:
First part of #1772

### DIFF
--- a/src/back/models/Personnel.ts
+++ b/src/back/models/Personnel.ts
@@ -10,13 +10,13 @@ export type PersonnelAttributes = {
   status: PersonnelStatus
   updated_by?: string
   updated_at?: Date
-  created_at?: Date
+  created_at: Date
 }
 
 export enum PersonnelStatus {
-  Deleted = 'deleted',
-  Unassigned = 'unassigned',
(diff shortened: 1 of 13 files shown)


src/back/models/Personnel.ts
```diff
@@ -0,0 +1,26 @@
...
+export enum PersonnelStatus {
+  Deleted = 'Deleted',
+  Unassigned = 'Unassigned',
+  Assigned = 'Assigned',
+}
+
+export default class PersonnelModel extends Model<PersonnelAttributes> {
+  static tableName = 'personnel'
+  static withTimestamps = false
+  static primaryKey = 'id'
```
ncomerci, line 25: `static primaryKey = 'id'`

Id is not within PersonnelAttributes 🤔

src/back/models/Personnel.ts
ncomerci, line 1: Perhaps we can think of a way in which the same personnel are assigned to different projects

ncomerci (approved):

src/back/models/Personnel.ts
1emu: Yeah, the thing is that the same person could have a different role on each project

Issue #1763, opened 2024-04-24
Passed Grant Proposals display a component taking the users to the Project Sheet


ginocingolani:

andyesp:
- What is the component status for a Done/Finished project?

PR #1786, opened 2024-05-08
chore: add basic project info to proposal data on proposals endpoint

1emu:
Part of #1763 

Change single proposal endpoint to include basic project attributes on the response, so we can link to the project from the proposal page

image

image


src/back/routes/proposal.ts
```diff
@@ -131,7 +131,7 @@ export default routes((route) => {
   route.post('/proposals/hiring', withAuth, handleAPI(createProposalHiring))
   route.get('/proposals/priority/:address?', handleJSON(getPriorityProposals))
   route.get('/proposals/grants/:address', handleAPI(getGrantsByUser))
-  route.get('/proposals/:proposal', handleAPI(getProposal))
+  route.get('/proposals/:proposal', handleAPI(getProposalWithProject))
```
ncomerci, line 134: I would not rename this function since these new attributes belong to the proposal

ncomerci (approved):

PR #1795, opened 2024-05-15
feat: get projects query

ncomerci:


src/back/models/Project.ts
```diff
@@ -15,8 +16,27 @@ export type ProjectAttributes = {
...
+}
+
 export default class ProjectModel extends Model<ProjectAttributes> {
   static tableName = 'projects'
   static withTimestamps = false
   static primaryKey = 'id'
+
+  static async getProject(id: string) {
+    const query = SQL`
+    SELECT pr.id, pr.proposal_id, p.user as author, array_agg(co.address) FILTER (WHERE co.status = 'APPROVED') as coauthors, pr.title, pr.status, pr.links, pr.about, pr.about_updated_by, pr.about_updated_at, pr.updated_at, pr.created_at
```
1emu, line 32: Nice! Since we are using all project data I think you could just `SELECT pr.*`

Also, we won't use any unapproved coauthors data in the query, so maybe you could just filter them in the `ON` clause of the join, before aggregating.

```sql
SELECT pr.*, p.user as author, array_agg(co.address) as coauthors
FROM projects pr
JOIN proposals p on pr.proposal_id = p.id
LEFT JOIN coauthors co ON pr.proposal_id = co.proposal_id AND co.status = 'APPROVED'
WHERE pr.id = ${id}
GROUP BY pr.id;
```

src/back/models/Project.ts
```diff
@@ -15,8 +16,27 @@ export type ProjectAttributes = {
...
+    FROM projects pr
+    JOIN proposals p on pr.proposal_id = p.id
+    LEFT JOIN coauthors co on pr.proposal_id = co.proposal_id
+    WHERE pr.id = ${id}
+    GROUP BY pr.id, pr.proposal_id, p.user, pr.title, pr.status, pr.links, pr.about, pr.about_updated_by, pr.about_updated_at, pr.updated_at, pr.created_at;
+    `
+    const result = await this.namedQuery<ProjectAugmented>(`get_project_${id}`, query)
```
1emu, line 39: Are you sure we want to include the id in the query name? 🤔 
If this is for profiling, maybe it's better to run all `get_project` queries under the same key

src/back/models/Project.ts
```diff
@@ -15,8 +16,27 @@ export type ProjectAttributes = {
   created_at: Date
 }
 
+// TODO: add here all data from other tables (updates, personnel, milestones, etc)
+export type ProjectAugmented = ProjectAttributes & {
```
1emu, line 20: WDYT about calling this `Project`? (I like this name though)

1emu (approved):

src/services/ProjectService.ts
```diff
@@ -178,7 +178,7 @@ export class ProjectService {
 
   static async getProject(id: string) {
     //TODO: add all data to project from other tables (updates, personnel, milestones, etc) & return Project type, instead of ProjectAttributes
-    const project = ProjectModel.findOne<ProjectAttributes>({ id })
+    const project = ProjectModel.getProject(id)
```
1emu, line 181: I think you need to `await`this

1emu (approved):

andyesp (approved):

PR #1796, opened 2024-05-15
chore: create project personnel

### LINKED ISSUE #1790: Projects Sheet: create project Personnel from proposal

1emu:
Closes #1790 

You'll need to run `npm run migrate down` x4 before going up 🚀


src/back/models/Project.ts
```diff
@@ -29,14 +35,39 @@ export default class ProjectModel extends Model<ProjectAttributes> {
...
+                             'created_at', pe.created_at
+                     ) ORDER BY pe.id) FILTER (WHERE pe.id IS NOT NULL), '[]') AS personnel,
+            COALESCE(array_agg(co.address) FILTER (WHERE co.address IS NOT NULL), '{}') AS coauthors
+        FROM ${table(ProjectModel)} pr
+                 JOIN ${table(ProposalModel)} p ON pr.proposal_id = p.id
+                 LEFT JOIN ${table(PersonnelModel)} pe ON pr.id = pe.project_id AND pe.deleted = false
+                 LEFT JOIN ${table(CoauthorModel)} co ON pr.proposal_id = co.proposal_id AND co.status = ${
+      CoauthorStatus.APPROVED
+    }
+        WHERE pr.id = ${id}
```
andyesp, line 62: Are we validating this is an id?

src/services/ProjectService.ts
```diff
@@ -164,20 +166,46 @@ export class ProjectService {
...
+
+    await ProjectService.createPersonnel(proposal, newProject, creationDate)
 
-    return await ProjectModel.create(newProject)
+    return newProject
+  }
+
+  private static async createPersonnel(
+    proposal: ProposalWithOutcome,
+    newProject: ProjectAttributes,
```
andyesp, line 186: This won't be always a new project, isn't it? Maybe we can call it `project` here.

src/services/ProjectService.ts
```diff
@@ -164,20 +166,46 @@ export class ProjectService {
...
+    })
+
+    await ProjectService.createPersonnel(proposal, newProject, creationDate)
 
-    return await ProjectModel.create(newProject)
+    return newProject
+  }
+
+  private static async createPersonnel(
+    proposal: ProposalWithOutcome,
```
andyesp, line 185: We don't we pass just `members` or personnel here?

src/back/models/Project.ts
1emu: We are in the route handler, but I'll add a check in the model for xtra safu

src/services/ProjectService.ts
1emu: Hmm I agree on the name change, but in this case we will always use a new project, since the `createProject` method is the only one calling this, and it uses the personnel from an existing proposal to create all the personnel for a new project.

If we want to create new personnel for a project that already exists, we should have another method that creates them one by one (this one uses `createMany`)

WDYT?

src/services/ProjectService.ts
1emu: I was looking at it as if finding/parsing the info needed to create the personnel was part of this method's responsibility (so `createProject` only handles the project attributes, then delegates the rest)

ncomerci (approved):

andyesp (approved):

PR #1826, opened 2024-05-31
chore: Projects migration

ncomerci:


andyesp (approved):

src/services/ProjectService.ts
```diff
@@ -166,6 +167,22 @@ export class ProjectService {
     })
   }
 
+  static async migrateProjects() {
+    const { data: projects } = await this.getProjects()
+    const migratedProjects = await ProjectModel.migrate(projects)
+
+    inBackground(async () => {
```
1emu, line 174: I'm not sure we want to do this in background. I think it'd be good to have more visibility on the outcome, and any possible unexpected errors. What about something like:

```typescript
 static async migrateProjects() {
    const migrationResult = { migratedProjects: 0, migrationsFinished: 0, migrationErrors: [] as string[], error: '' }
    try {
      const { data: projects } = await this.getProjects()
      const migratedProjects = await ProjectModel.migrate(projects)
      migrationResult.migratedProjects = migratedProjects.length

      for (const project of migratedProjects) {
        try {
          const proposal = await ProposalService.getProposal(project.proposal_id)
          await ProjectService.createMilestones(proposal, project, new Date(project.created_at))
          await ProjectService.createPersonnel(proposal, project, new Date(project.created_at))
          migrationResult.migrationsFinished++
        } catch (e) {
          migrationResult.migrationErrors.push(`Project ${project.id} failed with: ${e}`)
        }
      }

      await UpdateModel.setProjectIds()
      return migrationResult
    } catch (e) {
      const message = `Migration failed: ${e}`
      console.error(message)
      migrationResult.error = message
      return migrationResult
    }
  }
  ```

PR #1829, opened 2024-06-03
feat: add and delete milestone endpoints

andyesp:


src/back/routes/project.ts
```diff
@@ -129,3 +136,41 @@ async function deleteLink(req: WithAuth<Request<{ link_id: string }>>): Promise<
 
   return await ProjectService.deleteLink(link_id)
 }
+async function addMilestone(req: WithAuth): Promise<ProjectMilestone> {
+  const user = req.auth!
+  const { milestone } = req.body
+  validateId(milestone.project_id)
+  const project = await ProjectService.getProject(milestone.project_id)
```
1emu, line 143: Instead of 

```typescript
  validateId(milestone.project_id)
  const project = await ProjectService.getProject(milestone.project_id)
```
  
you can `await validateCanEditProject(user, milestone.project_id)`, which checks 

```typescript
async function validateCanEditProject(user: string, projectId: string) {
  validateId(projectId)
  validateAddress(user)
  const isValidEditor = await ProjectService.isAuthorOrCoauthor(user, projectId)
  if (!isValidEditor) {
    throw new RequestError("Only the project's authors and coauthors can edit the project", RequestError.Unauthorized)
  }
}
```

And you can avoid checking if the project exists because the author/coauthor query already does:

```typescript
static async isAuthorOrCoauthor(user: string, projectId: string): Promise<boolean> {
    if (!isUUID(projectId || '')) {
      throw new Error(`Invalid project id: "${projectId}"`)
    }
    if (!isEthereumAddress(user)) {
      throw new Error(`Invalid user: "${user}"`)
    }

    const query = SQL`
      SELECT EXISTS (
        SELECT 1
        FROM ${table(ProjectModel)} pr
        JOIN ${table(ProposalModel)} p ON pr.proposal_id = p.id
        LEFT JOIN ${table(CoauthorModel)} co ON pr.proposal_id = co.proposal_id 
              AND co.status = ${CoauthorStatus.APPROVED}
        WHERE pr.id = ${projectId}
          AND (p.user = ${user} OR co.address = ${user})
      ) AS "exists"
    `

    const result = await this.namedQuery<{ exists: boolean }>(`is_author_or_coauthor`, query)
    return result[0]?.exists || false
  }
  ```

File: src/back/routes/project.ts [hunk 2]
```diff
@@ -129,3 +136,41 @@ async function deleteLink(req: WithAuth<Request<{ link_id: string }>>): Promise<
...
+
+async function deleteMilestone(req: WithAuth<Request<{ milestone_id: string }>>): Promise<string | null> {
+  const user = req.auth!
+  const milestone_id = req.params.milestone_id
+  validateId(milestone_id)
+  const milestone = await ProjectMilestoneModel.findOne<ProjectMilestone>(milestone_id)
+  if (!milestone) {
+    throw new RequestError(`Milestone "${milestone_id}" not found`, RequestError.NotFound)
+  }
+  const project = await ProjectService.getProject(milestone.project_id)
```
1emu, line 168: I think you can avoid checking for the project since you use `validateCanEditProject`, and also if the milestone exists, then the `project_id` must exist cause its a fk

1emu (approved):

PR #1831, opened 2024-06-04
feat: milestones submit limit in proposals request

### LINKED ISSUE #1830: Grant/Bid Request: Limit milestones to 10

andyesp:
Closes #1830


src/entities/Bid/types.ts
```diff
@@ -104,6 +105,7 @@ export const BidRequestGeneralInfoSchema = {
       additionalProperties: false,
       required: [...Object.keys(MilestoneItemSchema)],
       properties: MilestoneItemSchema,
+      maxItems: MILESTONE_SUBMIT_LIMIT,
```
1emu, line 108: This is the max they can submit, but afterwards they could create more? Maybe we should also check for the limit in the addMilestone of the project service, after `validateCanEditProject`, something like `validateCanAddMilestone` which checks

`return MILESTONE_SUBMIT_LIMIT > ProjectMilestones.count({project_id})`

andyesp: We don't have a product decision regarding milestones in projects (not proposals).

andyesp: This is a limitation to also include milestones in the snapshot body.

PR #1835, opened 2024-06-05
refactor: Updates

ncomerci:


src/back/routes/update.ts
```diff
@@ -139,15 +145,11 @@ async function updateProposalUpdate(req: WithAuth<Request<{ update_id: string }>
...
-    proposal,
-    user!,
-    now,
-    isOnTime
+    user!
   )
 }
 
-async function deleteProposalUpdate(req: WithAuth<Request<{ update_id: string }>>) {
+async function deleteProjectUpdate(req: WithAuth<Request<{ update_id: string }>>) {
```
1emu, line 152: If we are using projects I'd rather use the project_id here too. (you could also use the same validation mentioned before, see deletePersonnel i.e.)

src/back/routes/update.ts
```diff
@@ -179,3 +181,89 @@ async function deleteProposalUpdate(req: WithAuth<Request<{ update_id: string }>
...
+}
+
+async function validateFinancialRecords(
+  project: Project,
+  financial_records: unknown
+): Promise<FinancialRecord[] | null> {
+  const { proposal_id, vesting_addresses } = project
+  const [vestingData, updates] = await Promise.all([
+    VestingService.getVestingInfo(vesting_addresses),
+    UpdateService.getAllByProposalId(proposal_id),
```
1emu, line 204: I'd be good to find all usages of this and switch to projects (I don't think we are going to show updates with proposals anymore right?)

src/back/routes/update.ts
```diff
@@ -179,3 +181,89 @@ async function deleteProposalUpdate(req: WithAuth<Request<{ update_id: string }>
...
+async function createProjectUpdate(req: WithAuth) {
+  const user = req.auth!
+  const { project_id, author, financial_records, ...body } = req.body
+  const { health, introduction, highlights, blockers, next_steps, additional_notes } = validate<UpdateGeneralSection>(
+    schema.compile(GeneralUpdateSectionSchema),
+    body
+  )
+  try {
+    const project = await ProjectService.getProject(project_id)
+    const isAuthorOrCoauthor = ProjectService.isAuthorOrCoauthor(user!, project.id)
```
1emu, line 223: I think the same validation can be used :)

src/back/services/discord.ts
```diff
@@ -218,7 +218,7 @@ export class DiscordService {
     if (DISCORD_SERVICE_ENABLED) {
       inBackground(async () => {
         try {
-          const publicUpdates = getPublicUpdates(await UpdateModel.find<UpdateAttributes>({ proposal_id: proposalId }))
+          const publicUpdates = getPublicUpdates(await UpdateService.getAllByProposalId(proposalId))
```
1emu, line 221: shouldn't we change this so newUpdate messages also uses projects instead of proposals?

src/back/services/update.ts
```diff
@@ -163,9 +177,9 @@ export class UpdateService {
     const update = await UpdateModel.createUpdate(data)
     try {
       if (financial_records) await FinancialService.createRecords(update.id, financial_records)
-      await DiscourseService.createUpdate(update, proposal.title)
-      await EventsService.updateCreated(update.id, proposal.id, proposal.title, user)
-      DiscordService.newUpdate(proposal.id, proposal.title, update.id, user)
+      await DiscourseService.createUpdate(update, title)
+      await EventsService.updateCreated(update.id, proposal_id, title, user)
```
1emu, line 181: This makes me think that if updates are created for projects, then there might be some event copies that we'll need to change

src/services/ProjectService.ts
```diff
@@ -319,6 +329,11 @@ export class ProjectService {
     return await ProjectModel.isAuthorOrCoauthor(user, projectId)
   }
 
+  static async findProjectByProposalId(proposal_id: string) {
```
1emu, line 332: Is this being used? 🤔

src/back/routes/update.ts
andyesp: I think we don't need the project_id. As we can get the update and check if the user can edit it with update.project_id

src/services/ProposalService.ts
```diff
@@ -295,9 +294,7 @@ export class ProposalService {
     await ProposalModel.update<ProposalAttributes>(update, { id })
 
     if (isEnactedStatus && isProject) {
-      const latestVesting = vesting_addresses![vesting_addresses!.length - 1]
-      const vestingContractData = await getVestingContractData(latestVesting, proposal.id)
-      await UpdateModel.createPendingUpdates(proposal.id, vestingContractData)
+      await UpdateService.initialize(proposal.project_id!, vesting_addresses)
```
andyesp, line 297: We shouldn't infer project_id exists here

src/back/services/discord.ts
ncomerci: yes! but I would do it in another pr :)

PR #1848, opened 2024-06-11
feat: Not editable project status

ncomerci:
Part of https://github.com/decentraland/governance-ui/issues/56#issue-2321233654

### DIFF
--- a/src/back/routes/update.ts
+++ b/src/back/routes/update.ts
@@ -31,7 +31,7 @@ import { VestingService } from '../../services/VestingService'
 import { ErrorCategory } from '../../utils/errorCategories'
 import type { Project } from '../models/Project'
 import { UpdateService } from '../services/update'
-import { validateCanEditProject } from '../utils/validations'
+import { validateIsAuthorOrCoauthor } from '../utils/validations'
 
 export default routes((route) => {
   const withAuth = auth()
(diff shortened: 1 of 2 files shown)


src/back/utils/validations.ts
```diff
@@ -184,11 +185,20 @@ export function validateStatusUpdate(proposal: ProposalAttributes, statusUpdate:
...
+  const isAuthorOrCoauthor = await ProjectService.isAuthorOrCoauthor(user, projectId)
+  if (!isAuthorOrCoauthor) {
+    throw new RequestError("User is not the project's author or coauthor", RequestError.Unauthorized)
+  }
+}
+
+export async function validateCanEditProject(user: string, projectId: string) {
+  await validateIsAuthorOrCoauthor(user, projectId)
+  const project = await ProjectService.getUpdatedProject(projectId)
+  const notEditableStatus = new Set([ProjectStatus.Finished, ProjectStatus.Revoked])
```
1emu, line 200: this could be a const outside the fn so the set isn't created each time

src/back/routes/update.ts
```diff
@@ -31,7 +31,7 @@ import { VestingService } from '../../services/VestingService'
 import { ErrorCategory } from '../../utils/errorCategories'
 import type { Project } from '../models/Project'
 import { UpdateService } from '../services/update'
-import { validateCanEditProject } from '../utils/validations'
+import { validateIsAuthorOrCoauthor } from '../utils/validations'
```
1emu, line 34: You are using the old fn :)

1emu (approved):

src/back/routes/update.ts
ncomerci: I had to divide the validation into 2 parts!

andyesp (approved):

PR #1865, opened 2024-06-19
fix: Restore forum post for updates

ncomerci:


src/back/jobs/UpdatesMissingForumPost.ts
```diff
@@ -0,0 +1,11 @@
+import UpdateModel from '../../entities/Updates/model'
+import { DiscourseService } from '../../services/DiscourseService'
+import { ProposalService } from '../../services/ProposalService'
+
+export async function restoreMissingUpdatesForumPost() {
+  const affectedUpdates = await UpdateModel.getUpdatesWithoutForumPost()
+  for (const update of affectedUpdates) {
+    const { title } = await ProposalService.getProposal(update.proposal_id)
+    await DiscourseService.createUpdate(update, title)
```
1emu, line 9: It'd be good to have some feedback (return value/log) so that we know all affected updates had their forum post created 👌

andyesp (approved):

Issue #1776, opened 2024-04-30
Projects Sheet: Activity


andyesp:
- Proposal created
- Proposal finished
- Vesting Contract enacted
- Project Update posted

Nice to have
Events from the vesting contracts
- Pause
- Revoke
- Resume
- Vest
- Release

andyesp:
* What events should we display in the activity tab?

andyesp:
* What about milestones completed or tasks done?

PR #1873, opened 2024-06-27
feat: Activity events filters

ncomerci:
Part of  https://github.com/decentraland/governance/issues/1776


src/models/Event.ts
```diff
@@ -20,32 +20,25 @@ export default class EventModel extends Model<Event> {
     return result
   }
 
-  static async getLatest(filters?: EventFilter): Promise<Event[]> {
-    const { event_type } = filters || {}
+  static async getLatest(filters: EventFilter): Promise<Event[]> {
+    const { withInterval, event_type, proposal_id } = filters
```
andyesp, line 24: We should use snake case here `with_interval`.

andyesp (approved):

PR #1886, opened 2024-08-05
chore: subgraph vesting data

1emu:
Data parsed from subgraph seems to match alchemy:

image

image

### DIFF
--- a/src/clients/VestingData.ts
+++ b/src/clients/VestingData.ts
@@ -3,14 +3,12 @@ import { JsonRpcProvider } from '@ethersproject/providers'
 import { ethers } from 'ethers'
 
 import { VestingStatus } from '../entities/Grant/types'
-import { CLIFF_PERIOD_IN_DAYS } from '../entities/Proposal/utils'
 import { ErrorService } from '../services/ErrorService'
 import RpcService from '../services/RpcService'
 import ERC20_ABI from '../utils/contracts/abi/ERC20.abi.json'
 import VESTING_ABI from '../utils/contracts/abi/vesting/vesting.json'
 import VESTING_V2_ABI from '../utils/contracts/abi/vesting/vesting_v2.json'
 import { ContractVersion, TopicsByVersion } from '../utils/contracts/vesting'
-import Time from '../utils/date/Time'
 import { ErrorCategory } from '../utils/errorCategories'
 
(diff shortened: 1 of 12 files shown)


src/services/VestingService.ts
```diff
@@ -18,9 +22,44 @@ export class VestingService {
   }
 
   static async getVestings(addresses: string[]): Promise<VestingWithLogs[]> {
-    const vestings = await Promise.all(addresses.map((address) => getVestingWithLogs(address)))
+    const vestingsData = await VestingsSubgraph.get().getVestings(addresses)
+    return vestingsData.map(this.parseSubgraphVesting).sort(compareVestingInfo)
```
andyesp, line 26: Unrelated but I think we should rename `compareVestingInfo` to something like `sortVestingsByDate`.

src/services/VestingService.ts
```diff
@@ -18,9 +22,44 @@ export class VestingService {
...
+
+  private static async getVestingWithLogsFromSubgraph(
+    vestingAddress: string,
+    proposalId?: string
+  ): Promise<VestingWithLogs> {
+    try {
+      const subgraphVesting = await VestingsSubgraph.get().getVesting(vestingAddress)
+      return this.parseSubgraphVesting(subgraphVesting)
+    } catch (error) {
+      console.log('Unable to fetch vestings subgraph data', error) //TODO: remove before merging to master
```
andyesp, line 48: Remove?

src/clients/VestingData.ts
```diff
@@ -172,26 +182,125 @@ async function getVestingContractDataV2(
...
   }
 }
 
-export async function getVestingWithLogs(
-  vestingAddress: string | null | undefined,
-  proposalId?: string
-): Promise<VestingWithLogs> {
-  if (!vestingAddress || vestingAddress.length === 0) {
-    throw new Error('Unable to fetch vesting data for empty contract address')
+export function parseVestingData(vestingData: SubgraphVesting): Vesting {
```
andyesp, line 199: We could move the parsing functions to another file like `utils/vesting.ts` as this file was focused on querying Vesting contract data directly from the contracts.

andyesp: Or move it to the VestingService file as it is not used anywhere else.

andyesp (approved):

Issue #1904, opened 2025-09-10
VP sources mismatched due to incorrect Snapshot strategy order


ginocingolani:
Description:
We currently retrieve VP sources (MANA, LAND, Delegation, NAMES, etc.) from the Snapshot API. After recent changes to the voting strategies in Snapshot, the sources are being displayed incorrectly.

Example: the UI shows that a user has X amount of MANA VP when in fact that X belongs to their LAND VP.

Cause:
	•	The governance platform assumes a fixed order of strategies when retrieving data from Snapshot.
	•	Snapshot has changed the order of the strategies in its API response.
	•	As a result, we are rendering mismatched values (i.e., mapping the wrong strategy to the wrong VP source).

Expected behavior:
Each VP source (MANA, LAND, Delegation, NAMES, etc.) should be correctly matched and displayed regardless of the order returned by Snapshot. This appears on the user profile and on the delegation view of the profile.

<img width="923" height="334" alt="Image" src="https://github.com/user-attachments/assets/d103aae6-7b5c-43b0-9c8b-3dc4746e2784" />

<img width="660" height="419" alt="Image" src="https://github.com/user-attachments/assets/6923d365-2348-42ed-8560-396759e8a33f" />

Snapshot voting strategies order

<img width="2978" height="3438" alt="Image" src="https://github.com/user-attachments/assets/ab5d7008-fd83-47cf-aca8-ba165f8b1e5d" />

douglasborthwick-crypto:
The underlying fragility here is depending on Snapshot strategy position — even after fixing the mapping, any Snapshot internals change could break it again.

An approach that removes that dependency: query each token type directly against a contract-keyed endpoint rather than relying on strategy ordering. `POST /v1/attest` from [InsumerAPI](https://insumermodel.com/developers/verification/) returns results labelled by the condition you define, not by array position:

```json
{
  "wallet": "0x...",
  "conditions": [
    { "type": "token_balance", "contractAddress": "0xMANA_CONTRACT", "chainId": 1, "threshold": 100, "decimals": 18, "label": "MANA" },
    { "type": "nft_ownership", "contractAddress": "0xLAND_CONTRACT", "chainId": 1, "label": "LAND" }
  ]
}
```

[...]

PR #1913, opened 2025-09-30
fix:missmatch-in-strategies

Gaffomdz:
**Screenshots**
before:
Captura de pantalla 2025-09-30 a las 16 38 02
after:
Captura de pantalla 2025-09-30 a las 16 40 46

**Related Issue**
https://github.com/decentraland/governance/issues/1904

in this PR:
Replaced enum-based mapping with explicit indices matching the space’s current order.

### DIFF
--- a/src/clients/SnapshotGraphql.ts
+++ b/src/clients/SnapshotGraphql.ts
@@ -19,6 +19,7 @@ import {
   SnapshotVote,
   SnapshotVoteResponse,
   SnapshotVpResponse,
+  StrategyOrder,
   VpDistribution,
 } from './SnapshotTypes'
 import { inBatches, trimLastForwardSlash } from './utils'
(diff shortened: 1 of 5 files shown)


src/clients/SnapshotGraphql.ts
```diff
@@ -351,48 +350,50 @@ export class SnapshotGraphql extends API {
...
+      json: { query, variables },
     })
 
-    if (!result?.data?.vp) {
-      throw Error('Unable to fetch VP Distribution')
-    }
+    if (!result?.data?.vp) throw Error('Unable to fetch VP Distribution')
+
+    const vpByStrategy = result.data.vp.vp_by_strategy
+    console.log('vpByStrategy', vpByStrategy)
```
gonpombo8, line 371: remove log

src/clients/SnapshotGraphql.ts
```diff
@@ -351,48 +350,50 @@ export class SnapshotGraphql extends API {
...
+    const total = Math.floor(result.data.vp.vp)
 
-    const vpByStrategy = result?.data?.vp.vp_by_strategy
+    const I = {
```
gonpombo8, line 374: Maybe instead of having here we can update the StrategyOrder enum with this values.
This way we ensure that we use the correct order in all the app

gonpombo8 (approved):

PR #1941, opened 2026-07-13
fix: security and correctness hardening from backend audit

LautaroPetaccio:
## What

Security and correctness hardening from a full backend audit. Grouped by severity.

### High
- **SSRF in `POST /url-title`** — now requires auth, resolves the host and rejects loopback/private/link-local/CGNAT/unique-local addresses (including alternate IPv6 spellings like `[0::1]`), and fetches with `redirect: 'manual'`.
- **Financial record delete/insert race** — `FinancialService.createRecords` now `await`s the delete before the insert so it can't wipe the just-inserted rows.

### Medium
- **`POST /budget/update`** gated behind a debug-address role.
- **Discourse update-comment dedup** — `isDiscourseEventRegistered` matches both `proposal_commented` and `project_update_commented`, so replays don't duplicate notifications.
- **Airdrop job** now `await`s all jobs with per-job error isolation (no unhandled rejection / premature "success").
- **Voted-event spoofing** — the event is only recorded after verifying the vote exists on Snapshot; title and choice are derived server-side (prevents activity-feed spoofing and stored XSS).
- **Subgraph errors** — new `queryGraphql` throws on non-2xx / GraphQL `errors`; `inBatches` reports failures and returns partial results instead of silently returning `[]`.
- **`updateProposalStatus`** runs the vesting work before persisting status, so a failure no longer leaves a half-enacted proposal.
- **Trusted-image bypass** — `extractImageUrls` now catches raw `` (quoted and unquoted) and strips markdown title suffixes; non-image hosts removed from the allowlist.
- **API-key leakage** — base client no longer follows redirects, so secret headers aren't re-sent to a redirect target.

### Low
- Request timeouts on all external client fetches.
- Re-enactment guard so an identical re-enact doesn't wipe/regenerate pending updates.
- Address casing/equality fixes (`isDAOCommittee`, proposal owner check, non-throwing `isSameAddress`).
- Consistent VP threshold in vote counting.
- `ProposalModel.update` WHERE clause joins with `AND` instead of `,`.
- `GET /debug` requires auth; `report-error` input validated/capped.
- Alchemy webhook propagates errors and rejects bad signatures.
- `validateBlockNumber` rejects `NaN`; newsletter body uses `JSON.stringify`; API keys redacted in error reports; webhook HMAC verified over the raw request body.

## Tests
Added unit tests for the SSRF IP checks (`routes/common.test.ts`), image-URL extraction and `validateBlockNumber` (`utils/validations.test.ts`), `redactSecrets` and `inBatches` (`clients/utils.test.ts`), the vote VP threshold boundary (`services/vote.test.ts`), and the re-enactment guard (`services/ProposalService.test.ts`).

## Test plan
- `npx tsc -p . --noEmit` — clean
- `npx jest` — 35 suites, 395 passed, 1 skipped

## Notes
- No endpoints were removed in this PR (that analysis is being tracked separately).
- A couple of behavior changes are deliberate: subgraph failures now surface/report instead of returning `[]` (partial results preserved), and API clients no longer follow redirects.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/routes/common.ts
+++ b/src/routes/common.ts
@@ -11,23 +11,6 @@ export default routes((router) => {
   return router.post('/url-title', withAuth, handleAPI(checkUrlTitle))
 })
 
-const fetchWithTimeout = async (url: string, timeout = 10000, options?: RequestInit) => {
-  const controller = new AbortController()
-  const timeoutId = setTimeout(() => {
-    controller.abort()
-  }, timeout)
-
-  try {
-    const response = await fetch(url, {
-      ...options,
-      signal: controller.signal,
-    })
-    return response
-  } finally {
-    clearTimeout(timeoutId)
-  }
-}
-
 const isHttpsURL = (url: string) => isURL(url, { protocols: ['https'], require_protocol: true })
 
 // Parses any valid IPv6 textual form (compact, expanded, or with an embedded IPv4 suffix)
@@ -107,6 +90,12 @@ export async function assertPublicUrl(url: string) {
     return
   }
 
+  // Known limitation: this resolves the hostname and vets the IPs, but the subsequent
+  // fetch() resolves DNS again independently, so a DNS-rebinding attacker (very low TTL)
+  // could pass here with a public IP and have the fetch hit a private IP. Fully closing it
+  // would require pinning this IP into the request (connect-to-IP + Host header), which
+  // native fetch doesn't support without breaking TLS SNI. Accepted for the current threat
+  // model (endpoint is authenticated); redirect: 'manual' in getTitle blocks the redirect variant.
   const resolved = await dns.promises.lookup(hostname, { all: true })
   if (resolved.length === 0) {
     throw new Error('Invalid url: could not resolve host')
@@ -117,12 +106,17 @@ export async function assertPublicUrl(url: string) {
 }
 
 async function getTitle(url: string) {
-  // redirect: 'manual' prevents a public URL from 3xx-redirecting to an internal host
-  // after the address check has passed.
-  const response = await fetchWithTimeout(url, 6000, { redirect: 'manual' })
-  const text = await response.text()
-  const title = text.match(/<title>([^<]+)<\/title>/)?.[1]
-  return title
+  const controller = new AbortController()
+  const timeoutId = setTimeout(() => controller.abort(), 6000)
+  try {
+    // redirect: 'manual' prevents a public URL from 3xx-redirecting to an internal host
+    // after the address check has passed.
+    const response = await fetch(url, { redirect: 'manual', signal: controller.signal })
+    const text = await response.text()
+    return text.match(/<title>([^<]+)<\/title>/)?.[1]
+  } finally {
+    clearTimeout(timeoutId)
+  }
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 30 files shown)


src/routes/common.ts
```diff
@@ -26,8 +30,96 @@ const fetchWithTimeout = async (url: string, timeout = 10000, options?: RequestI
...
+    if (a === 169 && b === 254) return true
+    if (a === 172 && b >= 16 && b <= 31) return true
+    if (a === 192 && b === 168) return true
+    if (a === 100 && b >= 64 && b <= 127) return true
+    return false
+  }
+
+  if (net.isIPv6(ip)) {
+    const bytes = parseIPv6ToBytes(ip.toLowerCase())
+    if (!bytes || bytes.length !== 16) return true // fail closed on anything unparseable
```
regenesis-claw, line 68: Observation (non-blocking): `assertPublicUrl` resolves DNS independently from the subsequent `fetch` call. A DNS rebinding attacker could return a public IP for this lookup and a private IP for the fetch's own resolution. The `redirect: 'manual'` flag closes the redirect-based variant, but the initial-request TOCTOU window remains. Fully closing it would require connecting to the resolved IP directly (with a `Host` header override). Current mitigation is reasonable for the threat model — just flagging for awareness.

src/services/events.ts
```diff
@@ -131,19 +133,39 @@ export class EventsService {
...
+      const choices: string[] = proposal.snapshot_proposal?.choices || []
+      const derivedChoice = typeof userVote.choice === 'number' ? choices[userVote.choice - 1] : undefined
+      const safeChoice = derivedChoice ?? (choices.includes(choice) ? choice : 'unknown')
+
       const votedEvent: VotedEvent = {
         id: crypto.randomUUID(),
         address,
         event_type: EventType.Voted,
-        event_data: { proposal_id, proposal_title, choice },
+        event_data: { proposal_id, proposal_title: proposal.title, choice: safeChoice },
```
regenesis-claw, line 162: Nice defense-in-depth: deriving the choice from the Snapshot vote index prevents stored XSS and feed spoofing, while the fallback to checking `choices.includes(choice)` handles edge cases where the vote schema differs. Clean approach.

regenesis-claw (approved):
Thorough security hardening — reviewed all 30 files. This is excellent work. The SSRF protection, vote spoofing fix, financial race condition, API key leak prevention, and webhook HMAC fixes are all well-implemented with good test coverage. A few observations (non-blocking):

**DNS rebinding TOCTOU (low risk, known limitation)**
`assertPublicUrl` resolves the hostname and checks the IPs, but `fetchWithTimeout` performs its own independent DNS resolution. An attacker with a DNS rebinding setup could pass the check with a public IP on the first resolution and have the actual fetch hit a private IP on the second. The `redirect: 'manual'` flag mitigates the redirect-based variant, but the initial-request rebinding window remains open. To fully close this, the resolved IP would need to be passed to `fetch` directly (e.g., connecting to the IP with a `Host` header override). This is a known hard problem and the current mitigation is reasonable for the threat model — just noting it for awareness.

**Two `fetchWithTimeout` functions with different signatures**
`routes/common.ts` has a local `fetchWithTimeout(url, timeout, options)` while `clients/utils.ts` exports `fetchWithTimeout(url, init, timeoutMs)`. Both work correctly in their own context, but if someone later imports the wrong one, the argument order swap would silently misapply the timeout. Consider aligning the signatures or renaming one (e.g., the local one in common.ts could just inline the AbortController since it's only used once).

**Webhook raw body fallback**
The `getWebhookPayload` fallback to `JSON.stringify(req.body)` is fine for current routes under `/api/webhooks`, but if a webhook handler is ever registered outside that path prefix, it won't get `rawBody` and the HMAC will verify against re-serialized JSON (which may differ from the sender's original payload due to key ordering, whitespace, etc.). A comment noting this coupling would help future contributors.

Overall: clean, well-tested, addresses real vulnerabilities. Approved.

Requested by Gon via Slack

PR #1945, opened 2026-07-28
test: cover the project routes and the ownership query behind them

LautaroPetaccio:
continues the coverage work from #1944, starting with the routes that carry
ownership checks. split along where the behaviour lives: anything the database
decides is covered against a real postgres, everything else is a unit test.

## the ownership query — integration
`ProjectModel.isAuthorOrCoauthor` is what every project write depends on, and it
is a single sql statement, so only a database can say what it returns. covered:
the author is allowed, an accepted coauthor is allowed, an invitation that is
still pending or was declined is not, a stranger is not, and an accepted coauthor
of a different project does not carry over. an unknown project is refused rather
than allowed, and a non-uuid project or non-address caller is refused before the
query runs.

## the routes — unit
what the routes add is the guard and the shape checks, so they run without a
database. every add and delete is covered for the unauthenticated caller, one who
may not edit, and the owner. the deletes read the owning project from the stored
row rather than the request, so a missing row is a 404 before ownership is
consulted and a non-uuid id never reaches the lookup. the editable-status half of
the guard is covered as well: a finished or revoked project refuses edits from
its own author.

removing either half of `validateCanEditProject` fails these — the status rule
three cases, the ownership check five.

## a finding, not fixed here
`ProjectModel.isAuthorOrCoauthor` compares addresses exactly:

```sql
AND (p.user = ${user} OR co.address = ${user})
```

while `getUserProjects` in the same file normalises both sides with `lower()`,
coauthor rows are always written lowercased, and `routes/coauthor.ts` lowercases
`req.auth` before matching — so that value is not reliably lowercase. a caller
whose authenticated address arrives checksummed would therefore be refused edit
access to a project they genuinely coauthor.

it fails closed, so it denies rather than grants, which is why this is a
follow-up rather than a fix here. two tests record the current behaviour and are
labelled as recording it; flipping them is a one-line change once the intended
semantics are confirmed.

## testing
752 unit and 169 integration, all passing. every case was checked by reverting
the behaviour it guards and confirming the test fails.

### DIFF
--- a/.github/workflows/pull_request_build.yml
+++ b/.github/workflows/pull_request_build.yml
@@ -67,6 +67,9 @@ jobs:
     env:
       NODE_ENV: test
       CONNECTION_STRING: postgres://postgres:postgres@localhost:5432/governance_test
+      # Not a secret: the bid payload column is encrypted at rest, and the round trip only runs if
+      # both halves use the same key.
+      DB_ENCRYPTION_KEY: integration-test-key
 
     steps:
       - uses: actions/checkout@v4
--- a/package.json
+++ b/package.json
@@ -31,7 +31,10 @@
       "^.+\\.(t|j)s$": "ts-jest"
     },
     "testEnvironment": "node",
-    "resolver": "<rootDir>/export-maps-resolver.js"
+    "resolver": "<rootDir>/export-maps-resolver.js",
+    "setupFilesAfterEnv": [
+      "<rootDir>/test/setup/closeTestApps.ts"
+    ]
   },
   "repository": {
     "type": "git",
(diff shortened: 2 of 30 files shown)


LautaroPetaccio:
on the second point about the webhook raw-body path: i had only fixed half of it
the first time, and the re-review was right to raise it again.

`2c8f754a` added a payload whose spacing and key order differ from what
`JSON.stringify` produces, signed over the bytes actually sent, but only for the
alchemy endpoint. discourse verifies its signature the same way and had the
identical gap — every case signed the re-serialised body and sent that same
object, so the captured bytes and the `JSON.stringify(req.body)` fallback were
byte-identical and neither path could be distinguished from the other.

`f9a6afbc` covers discourse the same way. the regression signal is what changed:
removing the raw-body capture from the test app used to fail four cases, and now
fails eight, across both endpoints.

worth stating plainly since it is the second time i reported a review comment as
handled when part of it was not. the first was the loose status assertions —
tightening the ones i had pattern-matched while missing the rest, which then
turned out to be hiding fourteen client errors reported as 500s.

PR #1954, opened 2026-08-10
fix: bound the proposal search term before it reaches the full-text parser

LautaroPetaccio:
## What

`GET /api/proposals` is unauthenticated, and its `search` term was only checked for holding two word characters — there was no length bound. The term reaches `pg-tsquery`, whose `word` pattern has three adjacent greedy quantifiers whose character classes all match `\s`, followed by a mandatory alternation. When the alternation cannot match, the engine enumerates every way of splitting a whitespace run across the three quantifiers, which is cubic.

Measured against the repo's own parser, with `search=ab|*` followed by N spaces (`+` in a query string decodes to a space, so one byte each, and `ab` satisfies the existing `\w{2}` check):

| spaces | time |
|--------|---------|
| 400 | 39 ms |
| 800 | 272 ms |
| 1600 | 2.1 s |
| 2400 | 6.9 s |
| 3200 | 16.4 s |

`getProposals` parses twice per request — once in `getProposalTotal`, once in `getProposalList`, both inside the same `Promise.all` — so a single ~3 kB anonymous GET stalled the entire single-process event loop for roughly 32 seconds. The request counter mounted in front of the route does not help here, because one request is the whole attack.

## How

Two changes. Either one closes the hole; both are here so that neither the route nor any future caller has to depend on the other.

- `getProposals` now rejects a term longer than `MAX_PROPOSAL_SEARCH_LENGTH` (100), mirroring how it already rejects one that is too short. No real title search comes close to the bound.
- The parser wrapper collapses whitespace runs before parsing. `to_tsquery` treats any run as a single separator, so this changes no result — `governance     grant` and `governance grant` both produce `governance:*&grant:*`. After it, 40,000 spaces parse in under a millisecond.

The second change is what protects the other caller: `tsquery()` is invoked from two model methods, and bounding only the route would leave anything added later exposed.

## Testing

- New `src/entities/Proposal/tsquery.test.ts` covers the collapsing (same output for one space and for a run) and asserts the pathological input parses well under a second.
- New cases in `src/routes/proposal.test.ts` assert an over-long term short-circuits without touching the model, and that a term exactly at the bound still queries.
- Full suite green: 67 suites, 1100 tests. `npm run build` clean.

### DIFF
--- a/src/entities/Proposal/tsquery.test.ts
+++ b/src/entities/Proposal/tsquery.test.ts
@@ -39,8 +39,10 @@ describe('tsquery', () => {
       elapsedMs = Date.now() - startedAt
     })
 
+    // Bounded loosely on purpose. Cubic growth puts this input in the hours before the fix, so a
+    // generous ceiling still separates the two behaviours and cannot be tripped by a slow runner.
     it('should parse without backtracking', () => {
-      expect(elapsedMs).toBeLessThan(1000)
+      expect(elapsedMs).toBeLessThan(10000)
     })
   })
 })
--- a/src/entities/Proposal/tsquery.test.ts
+++ b/src/entities/Proposal/tsquery.test.ts
@@ -1,0 +1,46 @@
+import tsquery from './tsquery'
+
+describe('tsquery', () => {
+  describe('when the term holds a single space between words', () => {
+    let result: string
+
+    beforeEach(() => {
+      result = tsquery('governance grant')
+    })
+
+    it('should prefix-match every word', () => {
+      expect(result).toBe('governance:*&grant:*')
+    })
+  })
+
+  describe('and the term holds a run of whitespace between the same words', () => {
+    let result: string
+
+    beforeEach(() => {
+      result = tsquery('governance     grant')
+    })
+
+    // Collapsing the run must not change what is searched for, otherwise the guard below would be
+    // trading a stall for wrong results.
+    it('should produce the same query as a single space', () => {
+      expect(result).toBe('governance:*&grant:*')
+    })
+  })
+
+  // The parser's word pattern has three adjacent quantifiers whose classes all match whitespace, so
+  // before the run was collapsed this input backtracked cubically: 3200 spaces took ~16s on the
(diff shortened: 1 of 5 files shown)


LautaroPetaccio:
Thanks — addressed the timing nit, and recorded a decision on the other two.

**Timing flakiness (P2): fixed** in `dd045931`. Raised the ceiling from 1s to 10s. Cubic growth puts 40k spaces in the *hours* before the fix, so a much looser bound still separates the two behaviours and can't be tripped by a slow runner. Loosening costs no signal here.

**Bound applied before whitespace normalization (P2): keeping as-is,** and I added a comment saying why so nobody "fixes" it later. Normalizing first would mean accepting an arbitrarily large body that happens to collapse to something short — on an unauthenticated endpoint the thing worth capping is what was *received*, not what it reduces to. The cost is that a legitimate 100+ character term made mostly of whitespace returns empty, which no real title search hits.

**Parser-level full input cap (P2): not doing it now,** deliberately. Both current callers reach the parser through this route's bound. A cap inside the wrapper would need a behaviour for exceeding it — throw or silently truncate — and neither is exercised by any caller today, so it'd be untested surface guarding a hypothetical. Worth adding *with* the caller that motivates it; the note in this thread is the pointer.

gonpombo8 (approved):
Reviewed the diff against the actual `pg-tsquery` source — the catastrophic-backtracking claim checks out, and the length cap runs before the parser on the only code path that passes `search` into the model (no other `getProposalList`/`getProposalTotal` caller passes a search term). Whitespace collapsing is semantically safe, including inside quoted phrases, and the boundary tests (100 accepted, 101 rejected) are exact.

One nit, non-blocking: the comment in `tsquery.ts` says the collapse removes the blowup "for every caller", but the vulnerable quantifier classes also match `*&+<:,|` — a long run of those isn't whitespace-normalized, so a hypothetical future unbounded caller would still have a (smaller, roughly quadratic) blowup. Not exploitable today since the route caps length first; worth a one-line comment tweak.

PR #1955, opened 2026-08-10
fix: fail closed when the land owner list cannot be read

LautaroPetaccio:
## What

`getLandOwnerAddresses` returned `[]` on any failure, and the LandOwner badge job derives revocation from *absence* from that list. With an empty list, every current holder falls into `unlistedUsersWithMintedOrReinstatedBadge` and comes back as `usersWithBadgesToRevoke`.

Nothing threw, so `getEligibleUsersForBadge` returned no `error`, the guard in `giveAndRevokeLandOwnerBadges` did not fire, and `revokeBadges` signed and broadcast a real `revokeBadge` transaction per holder with the raft owner key. The job runs daily at 02:30 in production.

Any of these was enough to trigger it: a request timeout, a DNS or TLS blip, a non-JSON response, or a single tile missing `owner` — `landOwner.owner.toLowerCase()` threw inside the same `try` and became the same empty list.

Recovery is not symmetric, which is what makes this worse than a transient outage. Reinstatement only covers holders who also voted on Snapshot, so any land owner who never voted stayed revoked permanently. It also burns MATIC per address.

## How

- `getLandOwnerAddresses` throws `LandOwnersUnavailableError` instead of returning an empty list. It also refuses a response carrying no tile data, or tiles with no owners among them, because those reach the caller as the same revoke-everyone shape.
- A tile without an `owner` is now skipped rather than throwing, so one malformed entry no longer discards the entire list.
- The request carries `AbortSignal.timeout`; it previously had none, unlike every other outbound call in the repo.
- `giveAndRevokeLandOwnerBadges` skips the run and reports it rather than proceeding on a list it could not read. Nothing is granted either — a partial read cannot be distinguished from a complete one, and the next scheduled run retries.

Scope check: LandOwner is the only badge whose job acts on `usersWithBadgesToRevoke`. The other caller of `getEligibleUsersForBadge` discards it, so no other badge could be mass-revoked this way.

## Also in here

The `trimmedId === ''` branch in `revokeBadges` and `reinstateBadges` recorded an `InvalidBadgeId` failure but had no `continue`, so it fell through and submitted the transaction anyway with an empty badge id, then could push a second result for the same entry. Both now `continue`. The third site with this shape already used an `else` and was correct — this brings the three into line.

## Deliberately not here

A ratio cap on the revocation batch (refuse if the revoke set exceeds some fraction of holders) would be belt-and-braces on top of this. I left it out because it needs `getEligibleUsersForBadge` to return the holder total, which ripples into its other caller, and the realistic failure modes are all covered by the guards above — the Land API is not paginated, and a truncated body fails JSON parsing rather than yielding a partial list. Worth doing separately if you want the extra margin.

## Testing

- New `src/entities/Badges/utils.test.ts`: owners deduped and lowercased on the happy path; a tile with no owner skipped rather than discarding the list; throws on request failure, on missing tile data, and on tiles with no owners; the request passes an abort signal.
- New cases in `src/services/BadgesService.test.ts`: an unreadable list revokes nothing and never classifies holders, and reports the skipped run — plus a happy-path case asserting a genuine former owner is still revoked, so failing closed did not quietly disable the job.
- Full suite green: 67 suites, 1104 tests. `npm run build` clean.

### DIFF
--- a/src/entities/Badges/utils.test.ts
+++ b/src/entities/Badges/utils.test.ts
@@ -1,5 +1,3 @@
-import { ErrorClient } from '../../clients/ErrorClient'
-
 import { LandOwnersUnavailableError, getLandOwnerAddresses } from './utils'
 
 const FIRST_OWNER = '0x56d0B5eD3D525332F00C9BC938f93598ab16AAA7'
@@ -7,10 +5,6 @@ const SECOND_OWNER = '0x49E4DbfF86a2E5DA27c540c9A9E8D2C3726E278F'
 
 describe('getLandOwnerAddresses', () => {
   let fetchMock: jest.SpyInstance
-
-  beforeEach(() => {
-    jest.spyOn(ErrorClient, 'report').mockImplementation(() => undefined)
-  })
 
   afterEach(() => {
     jest.restoreAllMocks()
--- a/src/entities/Badges/utils.test.ts
+++ b/src/entities/Badges/utils.test.ts
@@ -1,0 +1,109 @@
+import { ErrorClient } from '../../clients/ErrorClient'
+
+import { LandOwnersUnavailableError, getLandOwnerAddresses } from './utils'
+
+const FIRST_OWNER = '0x56d0B5eD3D525332F00C9BC938f93598ab16AAA7'
+const SECOND_OWNER = '0x49E4DbfF86a2E5DA27c540c9A9E8D2C3726E278F'
+
+describe('getLandOwnerAddresses', () => {
+  let fetchMock: jest.SpyInstance
+
+  beforeEach(() => {
+    jest.spyOn(ErrorClient, 'report').mockImplementation(() => undefined)
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  describe('when the land api answers with owned tiles', () => {
+    let result: string[]
+
+    beforeEach(async () => {
+      fetchMock = jest.spyOn(globalThis, 'fetch').mockResolvedValue({
+        json: async () => ({
+          data: {
+            '0,0': { owner: FIRST_OWNER },
+            '0,1': { owner: FIRST_OWNER },
+            '1,0': { owner: SECOND_OWNER },
+          },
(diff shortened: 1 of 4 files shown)


LautaroPetaccio:
Both P2s were real — fixed in `f56ee009`.

**Gas estimated against a possibly-invalid first id.** Good catch, and it was worse than a batch failure: the estimate throws *before* the loop records anything, so one bad id at the front failed every revocation in the batch, with no per-item results to show which. Both `revokeBadges` and `reinstateBadges` now estimate against the first id that trims to something, and when no id in the batch is usable they answer with a per-recipient `InvalidBadgeId` result rather than attempting an estimate or a transaction at all.

Covered with a `describe.each` over both methods: all-unusable ids never estimate and never submit; and — the case that was actually broken — a *leading* unusable id no longer stops the usable one behind it from being acted on.

**Duplicate error report.** Dropped the `ErrorClient.report` inside `getLandOwnerAddresses`. The caller's report is the more useful of the two (it names the skipped run and carries the cause), and the throw already propagates everything. `ErrorClient` and `ErrorCategory` are no longer used in that module, so their imports are gone too.

Full suite green: 67 suites, 1116 tests.

gonpombo8 (approved):
Verified the mass-revocation bug is real on master (fetch failure → `[]` → every holder classified for revocation) and that the fix closes it: `getLandOwnerAddresses` has exactly one call site, which now catches `LandOwnersUnavailableError`, reports it, and skips the run — no unhandled rejection can reach the cron runner or the debug route. The `continue` fixes in `revokeBadges`/`reinstateBadges` are correctly placed before the on-chain call, and `performBadgeAction` was indeed already correct. The 30s timeout matches the repo's other outbound-call defaults. Test coverage exercises the exact failure modes.

Two stylistic notes, non-blocking: `AbortSignal.timeout` is used directly rather than the existing `fetchWithTimeout` helper in `src/clients/utils.ts`, and the error interpolates its cause into the message instead of using the ES2022 `cause` option. Both functionally fine.

PR #1956, opened 2026-08-10
fix: gate proposal removal on the proposal status

LautaroPetaccio:
## What

`validateRemoval` checked only that the caller was the author or on the council. There was no status gate, so **any proposal author could delete their proposal in any status** — including `passed` and `enacted`.

That matters more than a stray DB row, because `removeProposal` does three things:

1. marks the proposal `deleted`,
2. drops the Discourse topic, and
3. sends a Snapshot `cancelProposal` signed with the DAO's own key.

So the author of an enacted grant that is being paid out by a live vesting contract could erase the proposal, the forum thread, **and** the Snapshot vote record — leaving the payout with nothing on record authorising it. Secondary effects: the category budget stays `allocated` forever with nothing pointing at it, and a tender author who is losing can delete their tender and submit a fresh one for a new window, since the tender-window checks read a deleted-filtered query.

`isProposalDeletable` already encodes exactly the intended rule — `active` or `pending` only — but grep showed its only references were in its own unit test. It was dead code.

## How

`isProposalDeletable` is now wired into the author path of `validateRemoval`.

The council path is deliberately unchanged: council members keep the ability to remove a proposal in any status, so this closes the author hole without removing a capability anyone may be relying on. If you'd rather the council also be constrained, that's a follow-up decision rather than something to fold in here.

A wrong-status removal answers `400`, matching how `validateStatusUpdate` already reports an invalid status transition.

## Testing

New `removeProposal` cases in `src/services/ProposalService.test.ts`:

- the author can still remove an `active` or `pending` proposal;
- the author is refused on `passed`, `enacted`, `rejected` and `finished`, and in each case the proposal is not marked deleted, the forum topic is not dropped, and the Snapshot proposal is not cancelled;
- a council member can still remove an `enacted` proposal, so the gate did not narrow the council path;
- a caller who is neither author nor council still gets `Forbidden`.

Full suite green: 66 suites, 1113 tests. `npm run build` clean.

### DIFF
--- a/src/entities/Proposal/markAsDeleted.test.ts
+++ b/src/entities/Proposal/markAsDeleted.test.ts
@@ -1,0 +1,97 @@
+import ProposalModel from './model'
+import { ProposalStatus } from './types'
+import { DELETABLE_PROPOSAL_STATUSES } from './utils'
+
+const PROPOSAL_ID = '00000000-0000-0000-0000-000000000001'
+const AUTHOR = '0x56d0b5ed3d525332f00c9bc938f93598ab16aaa7'
+const UPDATED_AT = new Date('2026-01-01T00:00:00.000Z')
+
+describe('ProposalModel.markAsDeleted', () => {
+  let namedRowCount: jest.SpyInstance
+  let queryText: string
+  let queryValues: unknown[]
+
+  function captureQuery(rowCount: number) {
+    namedRowCount = jest.spyOn(ProposalModel, 'namedRowCount').mockImplementation(async (_name, query) => {
+      queryText = query.text.replace(/\s+/g, ' ').trim()
+      queryValues = query.values
+      return rowCount
+    })
+  }
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  // The caller authorised the removal against a status it read earlier, so the same rule has to hold
+  // as a condition on the write: a job can move the proposal to passed or enacted in between.
+  describe('when the allowed statuses are given, as they are for the author path', () => {
+    beforeEach(async () => {
+      captureQuery(1)
+      await ProposalModel.markAsDeleted(PROPOSAL_ID, AUTHOR, UPDATED_AT, DELETABLE_PROPOSAL_STATUSES)
+    })
+
+    it('should constrain the update to those statuses', () => {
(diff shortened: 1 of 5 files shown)


LautaroPetaccio:
You're right, and this was the real gap — fixed in `9ab18ff8`. The status gate I added was a read check on a proposal the route had already fetched, so it made the rule *stated* but not *atomic*.

**What changed.** The removable statuses are now a condition on the write. `ProposalModel.markAsDeleted` runs `UPDATE … WHERE id = $1 AND deleted = FALSE AND status IN ($5, $6)` and answers whether a row was actually removed. `removeProposal` drops the Discourse topic and cancels the Snapshot proposal **only** on a true answer, and refuses with a 400 otherwise — so a proposal that moved to `passed` mid-request keeps its artefacts.

Two things beyond the report:

- The write also requires `deleted = FALSE`, so a repeated or concurrent request can't fire the external side effects twice. That was reachable before independently of the status race.
- `DELETABLE_PROPOSAL_STATUSES` is now exported and `isProposalDeletable` derives from it, so the read check and the write condition can't drift apart — which is the failure mode that would quietly reintroduce this.

The council path passes no statuses and stays unconstrained, as before.

**Testing.** New `markAsDeleted` tests assert against the generated SQL rather than a mocked call, so the `status IN` condition and the parameter binding are actually verified — plus that the council path emits no status condition, and that a zero-row update reports false. On the service side: the removal is refused and neither the forum topic nor the Snapshot proposal is touched when the write matches nothing, and both still fire when it succeeds.

I also printed both generated queries and checked the conditional doesn't leave dangling syntax on the council path. Full suite green: 67 suites, 1128 tests.

gonpombo8 (approved):
Verified end to end. `markAsDeleted` is reachable only via `DELETE /proposals/:proposal` → `removeProposal` → `validateRemoval`, the deletable-status list matches the enum exactly, and the statuses are bound as SQL parameters.

Worth highlighting: the PR is stronger than a plain validation fix — the status restriction is a condition on the `UPDATE ... WHERE status IN (...)` itself, and the Discourse-drop and Snapshot-cancel side effects only fire when the write matched a row. That atomically closes the check-then-act race with the finish job, so an author request that loses the race can no longer drop the forum topic or cancel the Snapshot vote. Council path (including a council member who is also the author) confirmed unchanged and tested.

PR #1957, opened 2026-08-10
fix: elect a tender or bid winner only once

LautaroPetaccio:
## What

`getFinishableLinkedProposals` expanded a finishable tender or bid into every sibling sharing its `linked_proposal_id` by calling `getProposalList({ type, linkedProposalId })` — with **no status**. The status filter in `getProposalList` is `conditional(!!status, …)`, so it was omitted entirely and siblings an earlier run had already decided came back into the election.

That matters because scores are fetched one proposal at a time and `calculateVotingResult` swallows a failure by returning `undefined`, so a group can be judged from a subset:

1. Bids B1…Bn for tender T all become finishable together (they share `publish_at`, hence `finish_at`).
2. Run 1: the score fetch fails for B2. B2 is dropped, the winner is elected among the rest → **B1 passes**, the others are rejected, B2 stays active.
3. Run 2: B2 is finishable. The expansion pulls the whole group back, including the already-passed B1. Scores all resolve. If B2 has more voting power, **B2 passes too**.
4. The corrective rejection for B1 matches zero rows, because `getFinishProposalQuery` is CAS'd on `status = 'active'`.

The tender now has two passed proposals, and `ProjectService.createProjects` filters on `newStatus === Passed`, so it creates a **second funded project** — two vesting contracts from one tender. The advisory lock does not help: these are sequential runs, not overlapping ones.

This is not hypothetical: the codebase already ships `reportPendingProposalsWithoutVotingResults` precisely because these per-proposal score fetches are observed to fail in production.

## How

- Only **active** siblings stand in the election.
- `getLinkedProposalGroups` also reports whether a group already holds a `passed` or `enacted` proposal. A straggler in an already-decided group is **rejected** rather than elected. This part matters: a status filter *alone* would have introduced a new variant of the same bug, because the straggler would then be the only candidate in its group and would win unopposed.
- A group whose members did not all resolve to a voting result is **deferred to the next run** rather than decided from a subset, since crowning the wrong proposal cannot be undone once the real winner's rejection is CAS'd away.

A group where every sibling was `rejected` is *not* treated as decided — nothing was elected there, so a remaining sibling may still legitimately win.

`getFinishableLinkedProposals` keeps its signature and delegates to the new function, so its existing coverage is untouched.

## Testing

New `getLinkedProposalGroups` cases: all-active groups return every candidate and are not flagged decided; a `passed` or `enacted` sibling is excluded from candidates and flags the group decided; an all-`rejected` group is not flagged decided.

New `finishProposal` cases driving the real job: a bid that becomes finishable after its group already elected a winner is rejected and never passed even though its voting power is the highest among active bids; a group with one unresolved voting result is not decided at all; and a fully-resolved group still passes the highest and rejects the rest.

I checked these tests actually catch the bug rather than just documenting the fix — emulating the old behaviour (drop the status filter, never flag a decided group) makes four of them fail, including the two-winners case.

Full suite green: 66 suites, 1104 tests. `npm run build` clean.

### DIFF
--- a/src/entities/Proposal/jobs.test.ts
+++ b/src/entities/Proposal/jobs.test.ts
@@ -11,7 +11,7 @@ import CoauthorModel from '../Coauthor/model'
 import { NewGrantCategory } from '../Grant/types'
 import { getQuarterEndDate } from '../QuarterBudget/utils'
 
-import { finishProposal, getFinishableLinkedProposals } from './jobs'
+import { finishProposal, getFinishableLinkedProposals, getLinkedProposalGroups } from './jobs'
 import ProposalModel from './model'
 import * as calculateOutcome from './outcome'
 import {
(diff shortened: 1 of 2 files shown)


LautaroPetaccio:
No code changes here — both P2s are things I'd rather leave visible than paper over. Reasoning for the record:

**DB-level single-winner invariant (P2).** Agreed this is the stronger guarantee, and it's the right eventual shape: a partial unique index over `(type, configuration->>'linked_proposal_id')` for non-deleted `passed`/`enacted` rows. I left it out because it's a migration against existing data, and if any tender already has two passed proposals from this bug the index creation fails — so it needs a data audit first, and that shouldn't ride along with the behavioural fix. Worth its own PR.

**`reportPendingProposalsWithoutVotingResults` firing for deferred groups (P2).** This one I think is correct as-is rather than noise to suppress. A deferral only happens *because* a score fetch failed, and that report is the signal saying so. Silencing it would leave a group quietly deferring every minute with nothing explaining why. The deferral itself logs separately, so the two together read as "scores failed → group waited", which is the story you want at 3am.

**In-memory status filtering (P2, first review).** The query has to return all statuses — detecting an already-decided group is the point of it, and that's what stops the straggler-wins-unopposed variant. Filtering after the fetch is inherent to that, not an oversight.

gonpombo8 (approved):
Confirmed the two-winners bug is real on master: the sibling expansion had no status filter, `calculateVotingResult` swallows score-fetch failures as `undefined`, and the CAS on `status = 'active'` means an already-passed winner could never be corrected — so a straggler could yield a second passed bid and a second funded project.

The fix checks out: a straggler in a decided group can never reach winner selection (the decided-check rejects it first — verified by the test where it holds the highest voting power), deferral holds back the whole group with nothing written so the retry is genuine, and the straggler's rejection flows through the standard notification/Discourse/Discord path. `getFinishableLinkedProposals` has no other callers, and the query pattern is unchanged from master.

One operational note, not a defect: a permanently failing score fetch for one sibling defers its whole group indefinitely, blocking the otherwise-resolved winner's funding. Intentional correctness-over-liveness tradeoff, and `reportPendingProposalsWithoutVotingResults` already alerts on it — just worth knowing if a group ever looks stuck.

